PR #21, opened 2017-08-02
[WIP] starting implementation of MOI interface

IssamT:
I didn't use MOIU for CLP since it's possible to call directly the solver methods, instead of building the whole model at `optimize!` calls. But now I think that, even doing so, it may be interesting to keep an MOIU instance in parallel to answer fast user queries. 

It's just the start but in case you have a comment on the approach, it might be efficient to rectify it early.

### DIFF
--- a/src/ClpSolverInterface.jl
+++ b/src/ClpSolverInterface.jl
@@ -255,6 +255,25 @@ immutable ClpMOISolver <: MOI.AbstractSolver
 end
 ClpMOISolver(;kwargs...) = ClpMOISolver(kwargs)
 
+function MOI.supportsproblem(s::ClpMOISolver, objective_type, constraint_types::Vector)
+    if objective_type != MOI.ScalarAffineFunction{Float64}
+        return false
+    end
+    supported_constraint_types = [(MOI.ScalarAffineFunction{Float64}, MOI.LessThan{Float64}),
+            (MOI.ScalarAffineFunction{Float64}, MOI.GreaterThan{Float64}),
+            (MOI.ScalarAffineFunction{Float64}, MOI.Interval{Float64}),
+            (MOI.SingleVariable, MOI.LessThan{Float64}),
+            (MOI.SingleVariable, MOI.GreaterThan{Float64})]
+    for v in constraint_types
+        if !(v in supported_constraint_types)
+            return false
+        end
+    end
+    return true
+end
+
+MOI.getattribute(s::ClpMOISolver, ::MOI.SupportsDuals) = true
+
 # Solver instance
 
 const VR = MOI.VariableReference
@@ -268,15 +287,25 @@ type ClpMOIModel <: MOI.AbstractSolverInstance
     vartoubref::Vector{CR}
     ubreftovar::Dict{CR, Int32}
     lbreftovar::Dict{CR, Int32}
+    nupperbounds::UInt64
+    nlowerbounds::UInt64
     nconstraints::UInt64 # does take into account bounds
     # constraints::Vector{CR}
-    nrows::UInt64 # does not take into account bounds
+    nrows_greaterthan::UInt64
+    nrows_lessthan::UInt64
+    nrows_interval::UInt64
+    nrows::UInt64
     rowtorangeref::Vector{CR}
     rangereftorow::Dict{CR, Int32}
+    solution::Vector{Float64}
+    rowsolution::Vector{Float64}
+    dualcolsolution::Vector{Float64}
+    dualrowsolution::Vector{Float64}
 end
 function ClpMOIModel(;kwargs...)
     m = ClpMOIModel(ClpModel(),ClpSolve(), 0, Vector{CR}(), Vector{CR}(), Dict{CR, Int32}(), Dict{CR, Int32}(),
-            0, 0, Vector{CR}(), Dict{CR, Int32}() )
+            0, 0, 0, 0, 0, 0, 0, Vector{CR}(), Dict{CR, Int32}(),
+            Vector{Float64}(), Vector{Float64}(), Vector{Float64}(), Vector{Float64}() )
     set_log_level(m.inner,0) # disable output by default
     for (name, value) in kwargs
         setoption(m, name, value)
@@ -285,7 +314,7 @@ function ClpMOIModel(;kwargs...)
 end
 
 col(m::ClpMOIModel,ref::VR) = ref.value - 1 # column index in solver
-row(m::ClpMOIModel,ref::CR) = rangereftorow[ref.value] - 1 # row index in solver
+row(m::ClpMOIModel,ref::CR) = m.rangereftorow[ref] - 1 # row index in solver
 
 MOI.SolverInstance(s::ClpMOISolver) = ClpMOIModel(;s.options...)
 
@@ -301,14 +330,14 @@ function MOI.addvariable!(m::ClpMOIModel)
     m.nconstraints += 2
     nv = m.nvariables
     nc = m.nconstraints
-    vf = VR(nv)
-    lbref = CR{MOI.ScalarVariablewiseFunction, MOI.LessThan{Float64}}(nc - 1)
-    ubref = CR{MOI.ScalarVariablewiseFunction, MOI.GreaterThan{Float64}}(nc)
+    vref = VR(nv)
+    lbref = CR{MOI.SingleVariable, MOI.GreaterThan{Float64}}(nc - 1)
+    ubref = CR{MOI.SingleVariable, MOI.LessThan{Float64}}(nc)
     push!(m.vartolbref, lbref)
     push!(m.vartoubref, ubref)
     m.lbreftovar[lbref] = nv
     m.ubreftovar[ubref] = nv
-    return vf
+    return vref
 end
 
 function MOI.addvariables!(m::ClpMOIModel, n::Integer)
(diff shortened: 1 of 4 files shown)


blegat:
Keeping an Instance with a copy of everything would be memory inefficient. Having fast accessors for `ConstraintFunction`, `ConstraintSet`, ... isn't such a priority.

src/ClpSolverInterface.jl
```diff
@@ -237,3 +237,195 @@ end
...
+        if x[i] > 1e20
+            x[i] = Inf
+        elseif x[i] < -1e20
+            x[i] = -Inf
+        end
+    end
+    return x
+end
+
+function MOI.addconstraint!(m::ClpMOIModel, f::MOI.ScalarVariablewiseFunction, s::MOI.GreaterThan{<:Real})
```
mlubin: This implementation has a quadratic cost for setting all variable bounds. This will definitely be too slow. The interface should keep its own copy of the bounds and pass them to Clp when needed.

src/ClpSolverInterface.jl
```diff
@@ -237,3 +237,195 @@ end
...
+
+type ClpMOIModel <: MOI.AbstractSolverInstance
+    inner::ClpModel
+    solveroptions::ClpSolve
+    nvariables::UInt64
+    vartolbref::Vector{CR}
+    vartoubref::Vector{CR}
+    ubreftovar::Dict{CR, Int32}
+    lbreftovar::Dict{CR, Int32}
+    nconstraints::UInt64 # does take into account bounds
```
mlubin: What is the purpose of counting bounds and linear constraints together?

src/ClpSolverInterface.jl
IssamT: I agree with you. We can keep it in the interface. We can also keep a boolean indicating if some element in the vector has changed since the last optimize.

src/ClpSolverInterface.jl
IssamT: It s not really needed. nconstraints can be also deduced as (2 * nvariables + nrows).

mlubin:
I'm a bit worried that adding constraints one-by-one will have a big performance hit compared to the current approach of providing the whole matrix at once. We'll need to run benchmarks. If this is the case, it would be a good idea for the solver wrapper to collect the linear constraints and give them to Clp all at once for the first solve. After the first solve, I imagine that we can just push all changes directly into the API.

CC @joaquimg

IssamT:
I don't know about CLP but I did have this issue with cplex in the past. grouping the modifications to send it once does make a difference, even after the first solve.

mlubin:
I'm most concerned about the first solve since slowing that down would constitute a performance regression compared with MPB. Past that I'm not too worried but you're welcome to optimize the interface if its worth your time.

joaquimg:
> I'm a bit worried that adding constraints one-by-one...
>  ...solver wrapper to collect the linear constraints and give them to Clp all at once for the first solve.

I think making the wrapper collect all constraints might increase a lot the complexity of the wrapper. Currently, I am in favor of the methods for vectors: `addconstraints!(m, f::Vector{ScalarAffineFunction}, s::Vector{GreaterThan})`. This way, a clever user (or someone upstream like JuMP) could use just 2 or 3 of those...

mlubin:
Oops, you're right. I forgot about that decision. The solver wrapper really shouldn't need to cache anything. We can use ``addconstraints!`` for specifying both blocks of bounds and blocks of linear constraints.

IssamT:
1) With `addconstraints!(m, f::Vector{ScalarAffineFunction}, s::Vector{GreaterThan})` solver wrapper still need to create the raw matrix and link indices to ConstraintReferences. it's just that the raw matrix will be a local variable instead of a cache.

2) This is fine for a user of JuMP, but a user of MOI will still suffer the slowdown when using `addconstraint`

joaquimg:
>  a user of MOI will still suffer the slowdown when using addconstraint

why?

mlubin:
> This is fine for a user of JuMP, but a user of MOI will still suffer the slowdown when using addconstraint

Users of MOI will have the choice of using ``addconstraint!`` or ``addconstraints!``, and each will have performance characteristics analogous to what would happen if you add constraints one-by-one or in chunks by using the solver's API directly. I don't see any problem with that.

IssamT:
because 1 addconstraints on an n size vector will be faster than n addconstraint if the solver wrapper is not caching.

Will JuMP use addconstraints for the whole constraint matrix or just for the set of constraints defined the macro @constraint. I am asking this because I have seen many people "misusing" the macro and adding anonymous single constraints within for loops.

IssamT:
I see your point @mlubin. comparing MOI to solver interfaces is a good argument.

joaquimg:
JuMP will use `addconstraints!` for all constraints of the same type that the user specified with `@constraint` before solve see [this comment](https://github.com/JuliaOpt/MathOptInterface.jl/issues/63#issuecomment-317526848).

joaquimg:
> I am asking this because I have seen many people "misusing" the macro and adding anonymous single constraints within for loops.

I am not sure that is a "misuse", I think it might be even faster to call `@constraint` inside a loop so that you may specify conditional statements in the right place.

IssamT:
Indeed it s not always a misuse.

odow:
I'd have a look at how we're structuring the CPLEX https://github.com/JuliaOpt/CPLEX.jl/pull/139 and Xpress https://github.com/JuliaOpt/Xpress.jl/pull/10 wrappers.

It seems to be the way to go (I haven't hit any real issues)...

IssamT:
similarly to `addconstraints!`, `deleteconstraints!` would be useful too

IssamT:
For implementing 

```julia
MOI.modifyconstraint!(m, c, ::MOI.ScalarCoefficientChange)
````
CLP_C_interface doesn t have a function to modify directly the coef of a column in a row, so I don't know if there is an alternative that avoids storing the constraint matrix in order to implement this as delete/add row(modified).

That's another point related to JuliaOpt/MathOptInterface.jl#106

**Comment by joaquimg** [comment]
Maybe CLP should not have the `modifyconstraint!` method for `ScalarCoefficientChange`. The user could use delete and add directly.
From what I understand the original idea was to keep it flexible so that MOI can do efficiently whatever the API can do efficiently. If the API cant change coefficients (has to delete and add to do it) then the MOI should do the same.

**Comment by IssamT** [comment]
I definitely agree with that since it goes along with what we have discussed before. wrappers should only implement MOI methods that are close to what the solvers offer.

**Comment by IssamT** [comment]
However, It would be easier if generic packages like JuMP always call the same functions. Utility functions like MOIU.modifyconstraint! are useful so that:

- In the mode where you have a cache: the modification would only be applied on the MOIU instance, and all the modifications will be pushed in batches using MOI supported functions once `MOIU.update_solver_instance!` is called.

- in the mode where you have no cache. either the solver supports `MOI.modifyconstraint!` and it will be called, or an error will be thrown otherwise.

**Comment by mlubin** [comment]
Note that ``ScalarCoefficientChange`` can be given a special meaning in the context of column generation. This is discussed at https://github.com/JuliaOpt/MathOptInterface.jl/issues/18.

**Comment by IssamT** [comment]
I have an issue with CLP status interpretation.

After solving this unbounded problem:
```
    #     min -x-y
    #     s.t. x-y == 0
    #     x,y >= 0
````

It returns:
```
primal_feasible(m.inner) = true
dual_feasible(m.inner) = true
is_proven_primal_infeasible(m.inner) = false
is_proven_dual_infeasible(m.inner) = true
```

Do you have an idea why `dual_feasible(m.inner)` is not false or should I ask on Coin mailing list?

**Comment by mlubin** [comment]
@IssamT, I took a quick look at the CLP source code, and I think the intention is for `is_proven_XXX` to have priority over `primal_feasible` and `dual_feasible`. The latter are statements about the current loaded solution and not the termination status.

File: src/ClpSolverInterface.jl [hunk 3]
```diff
@@ -643,23 +757,45 @@ function MOI.get(m::ClpMOIModel,
...
+    # @show is_proven_primal_infeasible(m.inner)
+    # @show is_proven_dual_infeasible(m.inner)
 end
 
 function MOI.get(m::ClpMOIModel, ::MOI.TerminationStatus)
     s = ClpCInterface.status(m.inner)
-    if s == 0
+    if s == 0 
         return MOI.Success
     elseif s == 1
```
mlubin: `InfeasibleNoResult` and `UnboundedNoResult` do not apply if a certificate is available. `s == 1` and `s == 2` should be `Success`.

mlubin:
More specifically, `is_proven_XXX` are just helper functions to interpret `status`: https://github.com/coin-or/Clp/blob/5443b4c9aff4416efb2cce2abe761a37e9ca9f22/Clp/src/ClpModel.hpp#L445

src/ClpSolverInterface.jl
IssamT: I agree with that. I actually have this commented in the code

```
# if s in [0, 1, 2] 
#     return MOI.Success
```

But I changed it because I can't find a method that returns the infeasibitliy certificate.

mlubin: https://github.com/JuliaOpt/Clp.jl/blob/33cb6c31e72bd958e409d3dcf868f1a8a16cec58/src/ClpCInterface.jl#L828

IssamT: Thanks

IssamT:
Closing this, and rewriting using the last version of MOI and LinQuadOptInterface

PR #27, opened 2018-05-29
WIP : MOI wrapper (based on LQOI)

IssamT:


src/MOIWrapper.jl
```diff
@@ -0,0 +1,627 @@
...
+
+"""
+get_linear_objective!(m, x::Vector{Float64})
+
+Change the linear coefficients of the objective and store
+in `x`.
+"""
+function LQOI.get_linear_objective!(instance::ClpOptimizer, x::Vector{Float64})
+    obj = get_obj_coefficients(instance.inner)
+    for i in 1:length(obj)
```
odow: `copy!(x, obj)`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,627 @@
...
+"""
+get_linear_constraint(m, row::Int)::Tuple{Vector{Int}, Vector{Float64}}
+
+Get the linear component of the constraint in the 1-indexed row `row` in
+the model `m`. Returns a tuple of `(cols, vals)`.
+"""
+function LQOI.get_linear_constraint(instance::ClpOptimizer, row::Int)::Tuple{Vector{Int}, Vector{Float64}}
+    A = get_constraint_matrix(instance.inner)
+    A_row = A[row,:]
+    return (Array{Int}(A_row.nzind) .- 1, A_row.nzval)
```
odow: With latest LQOI, we return 1-indexed columns

test/MOIWrapper.jl
```diff
@@ -0,0 +1,12 @@
...
+const MOI  = MathOptInterface
+const MOIT = MathOptInterface.Test
+
+@testset "Linear tests" begin
+    linconfig = MOIT.TestConfig(modify_lhs = false)
+    @testset "Default Solver"  begin
+        solver = ClpOptimizer(LogLevel = 0)
+        MOIT.contlineartest(solver, linconfig, ["linear10","linear12","linear8a","linear8b","linear8c"])
+    end
+end
```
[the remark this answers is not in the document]
IssamT: Added. Thanks for the review !

src/MOIWrapper.jl
```diff
@@ -0,0 +1,625 @@
...
+get_variable_primal_solution!(m, x::Vector{Float64})
+
+Get the primal solution for the variables in the model `m`, and
+store in `x`. `x`must have one element for each variable.
+"""
+function LQOI.get_variable_primal_solution!(instance::ClpOptimizer, x::Vector{Float64})
+    solution = primal_column_solution(instance.inner)
+    for i in 1:length(solution)
+        x[i] = solution[i]
+    end
```
odow: For all of these, there is `copy!(x, solution)`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,625 @@
...
+"""
+add_variables!(m, n::Int)::Void
+
+Add `n` new variables to the model `m`.
+"""
+function LQOI.add_variables!(instance::ClpOptimizer, n::Int)::Void
+    for i in 1:n
+        numberInColumn = 0
+        rows = Vector{Int32}()
+        elements = Vector{Float64}()
```
odow: I think `Int[]` and `Float64[]` are the preferred style over the above.

test/MOIWrapper.jl
```diff
@@ -0,0 +1,49 @@
...
+
+@testset "Unit Tests" begin
+    config = MOIT.TestConfig()
+    solver = ClpOptimizer(LogLevel = 0)
+
+    MOIT.basic_constraint_tests(solver, config, exclude = [
+        (MOI.SingleVariable, MOI.Integer),
+        (MOI.SingleVariable, MOI.ZeroOne)
```
odow: Hmm you shouldn't have to exclude these. This should only test constraints that you support. (See my comment in supported constraints.)

src/MOIWrapper.jl
```diff
@@ -0,0 +1,625 @@
...
+const SUPPORTED_CONSTRAINTS = [
+    (LQOI.Linear, LQOI.EQ),
+    (LQOI.Linear, LQOI.LE),
+    (LQOI.Linear, LQOI.GE),
+    (LQOI.SinVar, LQOI.EQ),
+    (LQOI.SinVar, LQOI.LE),
+    (LQOI.SinVar, LQOI.GE),
+    (LQOI.SinVar, LQOI.IV),
+    (LQOI.SinVar, MOI.ZeroOne),
+    (LQOI.SinVar, MOI.Integer),
```
odow: Clp doesn't support binary and integer constraints

test/MOIWrapper.jl
```diff
@@ -0,0 +1,49 @@
...
+    config = MOIT.TestConfig()
+    solver = ClpOptimizer(LogLevel = 0)
+
+    MOIT.basic_constraint_tests(solver, config, exclude = [
+        (MOI.SingleVariable, MOI.Integer),
+        (MOI.SingleVariable, MOI.ZeroOne)        
+    ])
+
+    MOIT.unittest(solver, config, [
+        "solve_affine_interval",
```
odow: Why is this excluded?

REQUIRE
```diff
@@ -1,3 +1,4 @@
 julia 0.6
 Cbc
 MathProgBase 0.5 0.8
+LinQuadOptInterface
```
odow: This should have a strict bound of `LinQuadOptInterface 0.1 0.2`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+end
+
+"""
+get_variable_lowerbound(m, col::Int)::Float64
+
+Get the lower bound of the variable in 1-indexed column `col` of the model `m`.
+"""
+function LQOI.get_variable_lowerbound(instance::ClpOptimizer, col::Int)::Float64
+    lower = replaceInf(get_col_lower(instance.inner))
+    return lower[col];
```
odow: Two comments. Since you only query one element, you don't need to call `replaceInf` on the whole lowerbound vector. But also, if `get_col_lower` returns `10+20`, why should we change that to `Inf`? The real bound in Clp is `10+20`.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+end
+
+"""
+get_variable_upperbound(m, col::Int)::Float64
+
+Get the upper bound of the variable in 1-indexed column `col` of the model `m`.
+"""
+function LQOI.get_variable_upperbound(instance::ClpOptimizer, col::Int)::Float64
+    upper = replaceInf(get_col_upper(instance.inner))
+    return upper[col];
```
odow: Ditto here. Also trailing `;` is unneeded.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+"""
+get_number_linear_constraints(m)::Int
+
+Get the number of linear constraints in the model `m`.
+"""
+function LQOI.get_number_linear_constraints(instance::ClpOptimizer)
+    return get_num_rows(instance.inner)
+end
+
+function addrow(row, lower, upper, instance, rows, cols, coefs)
```
odow: The ordering on these arguments is inconsistent with the rest of the calls. They could be typed for readability?

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+
+"""
+    add_linear_constraints!(m, A::CSRMatrix{Float64}, sense::Vector{Cchar}, rhs::Vector{Float64})::Void
+
+Adds linear constraints of the form `Ax (sense) rhs` to the model `m`.
+`sense` and `rhs` contain one element for each row in `A`.
+The `sense` is given by `backend_type(m, set)`.
+Ranged constraints (`set=MOI.Interval`) should be added via `add_ranged_constraint!` instead.
+See also: `LinQuadOptInterface.CSRMatrix`.
+"""
```
odow: Here and throughout: you don't need this docstring since it is already in LQOI

src/MOIWrapper.jl
odow: It is also pretty confusing that you pass in the A matrix but only add one row. At the very least it should be documented

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+            upper = rhs[row]
+        elseif sense[row] == Cchar('G')
+            lower = rhs[row]
+        elseif sense[row] == Cchar('E')
+            upper = lower = rhs[row]
+        else
+            error("sense must be Cchar(x) where x is in ['L','G',E']")
+        end
+        
+        addrow(row, lower, upper, instance, rows, cols, coefs)
```
odow: I got confused here, see comment with `addrow` above

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+    ub = upper[row]
+    
+    if lb > -Inf
+        lower[row] = coef
+        chg_row_lower(instance.inner, lower)
+    elseif ub < Inf
+        upper[row] = coef
+        chg_row_upper(instance.inner, upper)
+    else
+        error("Either row_lower or row_upper must be of abs less than 1e20")
```
odow: So when adding a constraint it allows bigger coefficients, but it truncates them to 1e20, but when modifying it doesn't? Confusing.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+
+"""
+set_linear_objective!(m, cols::Vector{Int}, coefs::Vector{Float64})::Void
+
+Set the linear component of the objective.
+"""
+function LQOI.set_linear_objective!(instance::ClpOptimizer, cols::Vector{Int}, coefs::Vector{Float64})::Void
+    obj_in = zeros(Float64, get_num_cols(instance.inner))
+    for i in 1:length(cols)
+        obj_in[cols[i]] = coefs[i]
```
odow: I think it is sufficient to go `obj_in[cols] .= coefs` instead of the loop

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+add_variables!(m, n::Int)::Void
+
+Add `n` new variables to the model `m`.
+"""
+function LQOI.add_variables!(instance::ClpOptimizer, n::Int)::Void
+    for i in 1:n
+        numberInColumn = 0
+        rows = Int32[]
+        elements = Float64[]
+        add_column(instance.inner, numberInColumn,  rows, elements, -Inf, +Inf, 0.0)
```
odow: So `Inf` bounds are okay here? Not `1e20`?

src/ClpCInterface.jl
```diff
@@ -418,12 +420,36 @@ function add_rows(model::ClpModel, number::Integer, row_lower::Vector{Float64},
     @clp_ccall addRows Void (Ptr{Void}, Int32, Ptr{Float64}, Ptr{Float64}, Ptr{Int32}, Ptr{Int32}, Ptr{Float64}) model.p number row_lower row_upper row_starts columns elements
 end
 
+#This function exists in cpp but not c interface
+function add_row(model::ClpModel, number_in_row::Integer, columns::Vector{Int32}, elements::Vector{Float64},
+        row_lower::Float64, row_upper::Float64)
+        _row_starts = Vector{Int32}(2)
```
mlubin: Naming: variables in the scope of a function shouldn't start with underscores. `_row_upper` could be `row_upper_vector`. Same comment below.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
+export ClpOptimizer
+
+using LinQuadOptInterface
+using Clp.ClpCInterface
+
+function replaceInf(x)
```
mlubin: naming: no camel case, use `replace_inf`.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+        optionmap[name](m.inner,value)
+    elseif haskey(solveoptionmap, name)
+        solveoptionmap[name](m.solveroptions,value)
+    else
+        error("Unrecognized option: $name")
+    end
+end
+
+function ClpOptimizer(;kwargs...)
+    # env = Env()
```
mlubin: Style: don't leave code commented without an explanation why. If the code doesn't do anything, remove it.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+get_number_linear_constraints(m)::Int
+
+Get the number of linear constraints in the model `m`.
+"""
+function LQOI.get_number_linear_constraints(instance::ClpOptimizer)
+    return get_num_rows(instance.inner)
+end
+
+"""
+Helper function for adding a row. it adds the row in the components
```
mlubin: Start docstrings with the function signature.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+"""
+function LQOI.get_number_linear_constraints(instance::ClpOptimizer)
+    return get_num_rows(instance.inner)
+end
+
+"""
+Helper function for adding a row. it adds the row in the components 
+that build the LQOI sparse matrix, and it adds the row in the solver.
+"""
+function addrow(row::Int, lower::Float64, upper::Float64, instance::ClpOptimizer,
```
mlubin: Naming: `add_row`.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+`sense` and `rhs` contain one element for each row in `A`.
+"""
+function LQOI.add_linear_constraints!(instance::ClpOptimizer, A::LQOI.CSRMatrix{Float64}, sense::Vector{Cchar}, 
+        rhs::Vector{Float64})::Void
+                        
+    rows = A.row_pointers
+    cols = A.columns
+    coefs = A.coefficients
+           
+    nbrows = length(rhs)
```
mlubin: Naming: `num_rows`.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+        return lb
+    elseif ub < Inf
+        return ub
+    else
+        error("Either row_lower or row_upper must be of abs less than 1e20")
+    end
+end
+
+"""
+get_linear_constraint(m, row::Int)::Tuple{Vector{Int}, Vector{Float64}}
```
mlubin: Formatting: indent the signature with 4 spaces so it displays like code.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+
+Get the linear component of the constraint in the 1-indexed row `row` in
+the model `m`. Returns a tuple of `(cols, vals)`.
```
mlubin: Which model `m`? There's no `m` below.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,558 @@
...
+"""
+function LQOI.get_termination_status(instance::ClpOptimizer)
+    s = ClpCInterface.status(instance.inner)
+    if s == 0 
+        return MOI.Success
+    elseif s == 1
+        return MOI.InfeasibleNoResult
+    elseif s == 2
+        return MOI.UnboundedNoResult
+    # if s in [0, 1, 2]
```
mlubin: Remove or explain why it's commented.

IssamT:
There is an error on Linux but not on Mac. Can someone with Linux check the origin of this error?

blegat:
Yes, linear11 is failing on my Linux computer. You can probably add it in the exclude list. The problem seems to come from Clp and not from the wrapper.

IssamT:
Thanks @blegat ! We had the same issue with Cbc (since it uses the C++ Clp ), so we disabled the linear11 test there as well.

mlubin:
@IssamT is this ready to merge?

mlubin:
LQOI 0.2 is released, could you update the PR? It's not good for the ecosystem to start being split by LQOI version.

IssamT:
Sorry @mlubin , but I think I am missing something here. 
- I have tested the package against MOI master and LQOI master.
- However Travis used LQOI 1.0 for the previous commit because the REQUIRE had a version upper bound for LQOI. 
- I removed LQOI version upper and made this new commit. Hence, travis used LQOI 2.0 to pass the tests.
Which updates of LQOI are you talking about?

mlubin:
LQOI 0.3 could easily contain a change that would break this package. There should be an upper bound on LQOI to prevent users from seeing a broken package, like we do with MOI and MPB dependencies.

mlubin:
The bounds on LQOI look good now. Clp was previously passing tests on 0.7 (https://travis-ci.org/JuliaOpt/Clp.jl/builds/364908514), so the failure looks like a regression. Could you take a quick look at why it's failing?

test/MOIWrapper.jl
```diff
@@ -0,0 +1,48 @@
...
+        solver2 = ClpOptimizer(LogLevel = 0)
+        MOIT.copytest(solver,solver2)
+    end
+end
+
+@testset "Unit Tests" begin
+    config = MOIT.TestConfig()
+    solver = ClpOptimizer(LogLevel = 0)
+    # TODO uncomment after adding MOIT.TestConfig.multiple_bounds
+    # MOIT.basic_constraint_tests(solver, config)
```
odow: See https://github.com/JuliaOpt/MathOptInterface.jl/pull/434

test/MOIWrapper.jl
```diff
@@ -0,0 +1,48 @@
...
+    MOIT.nametest(solver)
+    @testset "validtest" begin
+        MOIT.validtest(solver)
+    end
+    @testset "emptytest" begin
+        MOIT.emptytest(solver)
+    end
+    # TODO uncomment after adding MOIT.TestConfig.multiple_bounds
+    # @testset "orderedindicestest" begin
+    #     MOIT.orderedindicestest(solver)
```
odow: See https://github.com/JuliaOpt/MathOptInterface.jl/pull/434

src/ClpCInterface.jl
```diff
@@ -418,12 +420,36 @@ function add_rows(model::ClpModel, number::Integer, row_lower::Vector{Float64},
...
 
+#This function exists in cpp but not c interface
+function add_row(model::ClpModel, number_in_row::Integer, columns::Vector{Int32}, elements::Vector{Float64},
+        row_lower::Float64, row_upper::Float64)
+        row_starts_vector = Vector{Int32}(2)
+        row_starts_vector[1] = 0
+        row_starts_vector[2] = number_in_row
+        row_upper_vector = [row_upper]
+        row_lower_vector = [row_lower]
+        add_rows(model, 1, row_lower_vector, row_upper_vector, row_starts_vector, columns, elements)
```
odow: Isn't it just as simple to go
`add_rows(model, 1, [row_lower], [row_upper], [0, number_in_row], columns, elements)`

src/ClpCInterface.jl
```diff
@@ -418,12 +420,36 @@ function add_rows(model::ClpModel, number::Integer, row_lower::Vector{Float64},
...
+#This function exists in cpp but not c interface
+function add_column(model::ClpModel, number_in_column::Integer,  rows::Vector{Int32},
+                elements::Vector{Float64}, column_lower::Float64,
+                column_upper::Float64, objective::Float64)
+    _column_starts = Vector{Int32}(2)
+    _column_starts[1] = 0
+    _column_starts[2] = number_in_column
+    _column_upper = [column_upper]
+    _column_lower = [column_lower]
+    _objective = [objective]
```
odow: Ditto here, or at least no underscore variable names

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+    lowerbounds = replace_inf(get_col_lower(instance.inner))    
+    
+    for i in 1:length(senses)
+        if senses[i] == Cchar('U')
+            upperbounds[cols[i]] = values[i]
+        elseif senses[i] == Cchar('L')
+            lowerbounds[cols[i]] = values[i]
+        else            
+            error("sense is " * string(senses[i]) * ", but only " * Cchar('U') *  
+            " and " * Cchar('L') * " senses are supported")
```
odow: Is this preferred over string interpolation?
`"sense is $(senses[i]), but only 'U' and 'L' senses are supported."`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+    
+    chg_column_upper(instance.inner, upperbounds)
+    chg_column_lower(instance.inner, lowerbounds)
+end
+
+"""
+    get_variable_lowerbound(instance, col::Int)::Float64
+
+Get the lower bound of the variable in 1-indexed column `col` of the model `m`.
+"""
```
odow: These don't need to be documented since they are just copies of the documentation in LQOI.
https://github.com/JuliaOpt/LinQuadOptInterface.jl/blob/d6957f357227e692c0e15bd6c64e9d0a37b7b1f1/src/solver_interface.jl#L121-L125

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+function LQOI.get_number_linear_constraints(instance::ClpOptimizer)
+    return get_num_rows(instance.inner)
+end
+
+"""
+    append_row(row::Int, lower::Float64, upper::Float64, instance::ClpOptimizer, 
+                rows::Vector{Int}, cols::Vector{Int}, coefs::Vector{Float64})
+
+Helper function for adding a row. it adds the row in the components 
+that build the LQOI sparse matrix, and it adds the row in the solver.
```
odow: I'm confused by "it adds the row in the components that build the LQOI sparse matrix" since it doesn't appear to do that? Maybe something like:
```
Given a sparse matrix in the triplet-form `(rows, cols, coefs)`, 
add row `row` with upper bound `upper` and lower bound 
`lower` to the instance `instance`.
```
Also, why is `instance` not the first argument?

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+
+"""
+    get_rhs(instance, row::Int)::Float64
+
+Get the right-hand side of the linear constraint in the 1-indexed row `row` in
+the model `m`.
+"""
+function LQOI.get_rhs(instance::ClpOptimizer, row::Int)::Float64
+    lower = replace_inf(get_row_lower(instance.inner))
+    upper = replace_inf(get_row_upper(instance.inner))
```
odow: Why do we call `replace_inf` on all elements if we are only returning the `row`'th element?

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+
+"""
+    change_variable_bounds!(instance, cols::Vector{Int}, 
+        values::Vector{Float64}, senses::Vector)
+        
+Change the bounds of the variable.
+"""
+function LQOI.change_variable_bounds!(instance::ClpOptimizer, cols::Vector{Int}, 
+        values::Vector{Float64}, senses::Vector)
+
```
odow: I think a JuMP style-guide item will be to remove unnecessary whitespace like this.

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+"""
+    change_linear_constraint_sense!(instance, rows::Vector{Int}, sense::Vector{Cchar})::Void
+
+Change the sense of the linear constraints in `rows` to `sense`.
+"""
+function LQOI.change_linear_constraint_sense!(instance::ClpOptimizer, rows::Vector{Int}, sense::Vector{Cchar} )::Void
+    lower = replace_inf(get_row_lower(instance.inner))
+    upper = replace_inf(get_row_upper(instance.inner))
+        
+    for (i,row) in enumerate(rows)
```
odow: If you use a zip you don't have to access the `i`'th element. 

`for (sense, row) in zip(senses, rows)`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+    delete_rows(instance.inner, which)
+end
+
+
+"""
+    change_linear_constraint_sense!(instance, rows::Vector{Int}, sense::Vector{Cchar})::Void
+
+Change the sense of the linear constraints in `rows` to `sense`.
+"""
+function LQOI.change_linear_constraint_sense!(instance::ClpOptimizer, rows::Vector{Int}, sense::Vector{Cchar} )::Void
```
odow: Should be `senses`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+        ub = upper[row]
+        if lb > -Inf
+            rhs = lb
+        elseif ub < Inf
+            rhs = ub
+        else
+            error("Either row_lower or row_upper must be of abs less than 1e20")
+        end
+        
+        sense = sense[i]
```
odow: Bug? Is this tested? I shouldn't imagine that this passed...

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+
+
+"""
+    set_linear_objective!(instance, cols::Vector{Int}, coefs::Vector{Float64})::Void
+
+Set the linear component of the objective.
+"""
+function LQOI.set_linear_objective!(instance::ClpOptimizer, cols::Vector{Int}, coefs::Vector{Float64})::Void
+    obj_in = zeros(Float64, get_num_cols(instance.inner))
+    obj_in[cols] .= coefs
```
odow: Doesn't account for duplicates. See https://github.com/JuliaOpt/MathOptInterface.jl/pull/432

src/MOIWrapper.jl
odow: This can probably just be `add_column(instance.inner, 0, Int32[], Float64[], -Inf, Inf, 0.0)`

src/MOIWrapper.jl
```diff
@@ -0,0 +1,561 @@
...
+    lower = replace_inf(get_row_lower(instance.inner))
+    upper = replace_inf(get_row_upper(instance.inner))
+    lb = lower[row]
+    ub = upper[row]
+    
+    if lb > -Inf
+        return lb
+    elseif ub < Inf
+        return ub
+    else
```
odow: What about interval constraints? You need to implement [get_range](https://github.com/JuliaOpt/LinQuadOptInterface.jl/blob/d6957f357227e692c0e15bd6c64e9d0a37b7b1f1/src/solver_interface.jl#L198)

odow: And also [modify_ranged_constraints!](https://github.com/JuliaOpt/LinQuadOptInterface.jl/blob/d6957f357227e692c0e15bd6c64e9d0a37b7b1f1/src/solver_interface.jl#L182)

test/MOIWrapper.jl
```diff
@@ -0,0 +1,48 @@
...
+const MOIT = MathOptInterface.Test
+
+@testset "Linear tests" begin
+    linconfig = MOIT.TestConfig(modify_lhs = false)
+    @testset "Default Solver"  begin
+        solver = ClpOptimizer(LogLevel = 0)
+        # linear1 test is disabled due to the following bug.
+        # https://projects.coin-or.org/Clp/ticket/84
+        MOIT.contlineartest(solver, linconfig, ["linear1", "linear10","linear11",
+            "linear12","linear8a","linear8b","linear8c"])
```
odow: Why are these tests excluded?

odow:
@IssamT there is a PR with some changes: https://github.com/IssamT/Clp.jl/pull/1

IssamT:
@mlubin, looking quickly at the error on 0.7, apparently it is due to LQOI failing under 0.7, which is due to Nullables.jl failing under 0.7. 

Supporting LQOI on both 0.6, 0.7 requires Nullables.

joaquimg:
Do you know why we need nullables in LQOI?
I think its always good to avoid dependencies, but if there is no other way lets go for it.

IssamT:
`Nullable` is disappearing in julia 0.7 in favor of `Union{T, Nothing}`. Nullables.jl offer a generic way of using Nullables that hides the difference between 0.6 and 0.7 (see JuliaLang/julia#23642)

However, as far as I can tell, `Nullable` is only used for `single_obj_var::Nullable{MOI.VariableIndex}`

I am not sure if that justifies a dependency ...

odow:
Okay, Nullables was removed from LQOI. Now we just need to wait for MOI to be tagged, then LQOI, then we can merge this with a new bound for LQOI.

odow:
You just need to add 0.7 on travis and make sure it passes

test/MOIWrapper.jl
```diff
@@ -0,0 +1,57 @@
+using MathOptInterface
+
+const MOI  = MathOptInterface
+const MOIB = MathOptInterface.Bridges
+const MOIT = MathOptInterface.Test
+
+@testset "Unit Tests" begin
+    config = MOIT.TestConfig()
+    solver = Clp.Optimizer(LogLevel = 0)
+    # MOIT.basic_constraint_tests(solver, config)
```
odow: This can be enabled

IssamT:
Is there still something required for this PR?

odow:
There a lot of deprecation warnings, but they can be addressed in another PR.

PR #29, opened 2018-08-02
Deprec 07

matbesancon:
Lots of deprecations v0.(5?) -> v0.7
There are build errors because of WinRPM, somehow does not block Clp build and usage
Rest of warnings come from MPB when the package is called for tests.

### DIFF
--- a/src/Clp.jl
+++ b/src/Clp.jl
@@ -6,11 +6,13 @@ __precompile__()
 
 module Clp
 
+import Pkg
 
 include("ClpCInterface.jl")
 include("ClpSolverInterface.jl")
 
 using Clp.ClpMathProgSolverInterface
+
 export ClpSolver
 
 end # module
--- a/src/ClpCInterface.jl
+++ b/src/ClpCInterface.jl
@@ -3,6 +3,8 @@ module ClpCInterface
 # Load binary dependencies via Cbc package
 import Cbc
 
+using SparseArrays: SparseMatrixCSC
+
 export
     # Types
     CoinBigIndex,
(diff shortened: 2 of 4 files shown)


mlubin:
Thanks! Please fix the test failure on 0.6.

> Rest of warnings come from MPB when the package is called for tests.

We'd like to fix these as well.

matbesancon:
fixing 0.6 and 0.7 will be hairy if not impossible on the same PR, maybe this could remain a separate branch while 0.6 is kept around for JuliaOpt.

mlubin:
It's not that hairy or impossible, we've been doing it for all of the solver packages, MPB, and MOI so far. Usually it involves the `Compat` package. (e.g., `using Compat.Pkg` instead of `using Pkg`). Given that development for JuMP/MOI and the solver MOI wrappers is continuing on 0.6 for now, branching off the solvers based on 0.6/0.7 would create too much of a maintenance burden.

matbesancon:
Given the time-line for MOI and 0.7, MOI and JuMP 0.19 will come after 0.7 won't they? That's why I was thinking upgrading the current JuliaOpt ecosystem to 0.7 would be more imminent

mlubin:
Yes, JuMP 0.19 will be released after 0.7. I agree that updating the MPB ecosystem to 0.7 is important, but branching the packages off for 0.6 is not compatible with our current development process. We're developing the MOI and MPB interfaces in the same repositories. In my experience with other packages, it's not difficult to maintain compatibility.

matbesancon:
has the discussion in slack general changed any points in the JuliaOpt release process for the 1.0 transition?

mlubin:
The current plans are described at https://discourse.julialang.org/t/jump-and-julia-0-7. We will be continuing to maintain 0.6 compatibility on the solver packages.

matbesancon:
Can't 0.6-compatible be maintained in its own version? The package
repository handles the last 0.6 supported version, which can still be
patched to keep up with minor changes that could still be required

mlubin:
Mainstream JuMP/MOI development is continuing on Julia 0.6 and 0.7 simultaneously, so it's not about patching with minor changes. We would have to backport all updates to the MOI wrapper, which is much more work than using the `Compat` package to keep the package working on 0.6 and 0.7 at the same time.

PR #34, opened 2018-09-08
add an AppVeyor config

mlubin:


mlubin:
The failures on 0.7 are pretty sad, wasn't expecting those.
CC @odow

odow:
Just sitting in Houston on my way to Chicago, so won't get a change to look at this for a few days

mlubin:
Adding an allowed failure on 0.7. All the errors are with WinRPM, so hopefully this will be fixed in the switch to BB anyway.

PR #41, opened 2018-10-17
Add logo to the README

odow:


README.md
```diff
@@ -1,7 +1,10 @@
-Clp
-===
+# COIN-OR Linear Programming (Clp)
 
-Interface to the **[Clp]** linear programming solver. Provides a complete interface to the low-level C API, as well as an implementation of the solver-independent ``MathProgSolverInterface`` for efficiently solving sequences of linear programs.   
+![](https://www.coin-or.org/wordpress/wp-content/uploads/2014/08/COINOR.png)
```
mlubin: Does this logo imply that this package is part of COIN-OR?

odow: I'm not sure. The feeling is that users of Clp/Cbc/etc from Pyomo or JuMP aren't aware they are using COIN-OR stuff.

mlubin: If the COIN-OR folks are requesting the logo, then fine with me.

odow: So going forward COIN-OR needs to try a little harder to raise the profile of its brand. We've discussed a few things like:
- Add logos to external wrappers
- Be more explicit that "Clp" is "COIN-OR Linear Programming"
- Add a header like Ipopt has (hence our other discussion) on first start up saying something along the lines of "This is a COIN-OR project" with a link to the website. This header could go either in the C source, or in the Julia wrapper.
- Rename "Clp" to be "CoinLP" (or COINlp). Since this is pretty hard to do at the C level, it would be easier at the JuMP/Pyomo/NEOS level.

These points hold for Cbc which would become CoinBC (or COINbc, or CoinIP).

PR #44, opened 2018-10-24
Add mip starts. Closes #47

### LINKED ISSUE #47: Implement LQOI.add_mip_start
See https://github.com/JuliaOpt/JuMP.jl/issues/1554

IssamT:


src/MOIWrapper.jl
```diff
@@ -368,3 +368,13 @@ end
 function LQOI.delete_variables!(instance::Optimizer, start_col::Int, end_col::Int)
     delete_columns(instance.inner, [Cint(i-1) for i in start_col:end_col])
 end
+
+function LQOI.add_mip_starts!(instance::Optimizer, cols::Vector{Int},
+                              x::Vector{Float64})
+
+    input = zeros(Float64, get_num_cols(instance.inner))
+    for (col, val) in zip(cols, x)
```
blegat: If one starting value is not given, it is silently set to zero. This isn't what MOI documents.
Is an error already thrown in LQOI and cols is guaranteed to contain all variables ?

src/MOIWrapper.jl
```diff
@@ -368,3 +368,13 @@ end
 function LQOI.delete_variables!(instance::Optimizer, start_col::Int, end_col::Int)
     delete_columns(instance.inner, [Cint(i-1) for i in start_col:end_col])
 end
+
+function LQOI.add_mip_starts!(instance::Optimizer, cols::Vector{Int},
+                              x::Vector{Float64})
+
```
IssamT: Maybe I should check the size of cols before, since CLP doesn't support apparently partial initial solutions?

blegat: Yes, it would work if LQOI guarantees that there is no duplicates in the cols. Otherwise, you can use a BitSet

src/MOIWrapper.jl
```diff
@@ -368,3 +368,13 @@ end
...
+    input = zeros(Float64, get_num_cols(instance.inner))
+    for (col, val) in zip(cols, x)
+        input[col] = val
+    end
+    set_col_solution(instance.inner, input)
```
mlubin: Do you have any evidence that `Clp` is able to make use of a starting solution? I'm skeptical that it looks at these values. As far as I know, you have to provide a complete basis in order to make it hot start.

blegat:
@IssamT You can add "Closes https://github.com/JuliaOpt/Clp.jl/issues/47" in the description of the PR

ndinsmore:
I did some testing that seems to indicate that Clp will ignore even a complete solution if it is loaded.   SO for the time being should the pr be closed?

mlubin:
This is stale now that the MOI wrapper was rewritten without LQOI.

Issue #49, opened 2019-01-04
Update to LQOI 0.6


ianfiske:
Without this update, we can't grab the latest Cbc.jl (tagged) and Gurobi.jl (#master) due to conflicts:

```
(jtestenv) pkg> st
    Status `~/jtestenv/Project.toml`
  [9961bab8] Cbc v0.4.3
  [e2554f3b] Clp v0.6.0
  [2e9cd046] Gurobi v0.5.7

(jtestenv) pkg> add Cbc@0.5.0
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package LinQuadOptInterface [f8899e07]:
 LinQuadOptInterface [f8899e07] log:
 ├─possible versions are: [0.1.0, 0.2.0, 0.3.0-0.3.1, 0.4.0-0.4.3, 0.5.0, 0.6.0] or uninstalled
 ├─restricted by julia compatibility requirements to versions: [0.3.0-0.3.1, 0.4.0-0.4.3, 0.5.0, 0.6.0] or uninstalled
 ├─restricted by compatibility requirements with Clp [e2554f3b] to versions: 0.4.0-0.4.3
 │ └─Clp [e2554f3b] log:
 │   ├─possible versions are: [0.0.5-0.0.10, 0.1.0-0.1.1, 0.2.0-0.2.2, 0.3.0-0.3.1, 0.4.0, 0.5.0-0.5.1, 0.6.0] or uninstalled
 │   └─restricted to versions 0.6.0 by an explicit requirement, leaving only versions 0.6.0
 └─restricted by compatibility requirements with MathOptInterface [b8f27783] to versions: 0.6.0 or uninstalled — no versions left
   └─MathOptInterface [b8f27783] log:
     ├─possible versions are: [0.1.0, 0.2.0, 0.3.0, 0.4.0-0.4.1, 0.5.0-0.5.1, 0.6.0-0.6.4, 0.7.0, 0.8.0] or uninstalled
     ├─restricted by julia compatibility requirements to versions: [0.5.0-0.5.1, 0.6.0-0.6.4, 0.7.0, 0.8.0] or uninstalled
     └─restricted by compatibility requirements with Cbc [9961bab8] to versions: 0.8.0
       └─Cbc [9961bab8] log:
[...]

PR #50, opened 2019-01-04
Minimal updates for LQOI 0.6.0

ianfiske:
Addresses #49 .  I made the minimal changes necessary to work with LQOI 0.6.0 and have the tests pass.

### DIFF
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -89,6 +89,8 @@ function LQOI.get_objectivesense(model::Optimizer)
         return MOI.MIN_SENSE
     elseif s == -1.0
         return MOI.MAX_SENSE
+    elseif s == 0.0
+        return MOI.FEASIBILITY_SENSE
     else
         error("Invalid objective sense: $(s)")
     end
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -90,6 +90,7 @@ function LQOI.get_objectivesense(model::Optimizer)
     elseif s == -1.0
         return MOI.MAX_SENSE
     else
+        @assert iszero(s)
         return MOI.FEASIBILITY_SENSE
     end
 end
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -90,7 +90,7 @@ function LQOI.get_objectivesense(model::Optimizer)
     elseif s == -1.0
         return MOI.MAX_SENSE
     else
-        error("Invalid objective sense: $(sense)")
+        error("Invalid objective sense: $(s)")
     end
 end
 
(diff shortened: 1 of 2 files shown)


src/MOIWrapper.jl
```diff
@@ -82,6 +82,19 @@ function replace_inf(x::Real)
...
 
+
+function LQOI.get_objectivesense(model::Optimizer)
+    s = get_obj_sense(model.inner)
+    if s == 1.0
+        return MOI.MIN_SENSE
+    elseif s == -1.0
+        return MOI.MAX_SENSE
+    else
+        error("Invalid objective sense: $(sense)")
```
odow: `$(s)` not `$(sense)`.

ianfiske: good catch -- fixed.

src/MOIWrapper.jl
```diff
@@ -82,6 +82,19 @@ function replace_inf(x::Real)
...
 
+
+function LQOI.get_objectivesense(model::Optimizer)
+    s = get_obj_sense(model.inner)
+    if s == 1.0
+        return MOI.MIN_SENSE
+    elseif s == -1.0
+        return MOI.MAX_SENSE
+    else
+        error("Invalid objective sense: $(s)")
```
odow: Is it possible for `get_obj_sense` to return something else? Can we add a test?

blegat: Couldn't it be `MOI.FEASIBILITY_SENSE` ?

mlubin: According to https://projects.coin-or.org/Clp/browser/trunk/Clp/src/Clp_C_Interface.h#L419, it looks like a return value of 0.0 maps to `MOI.FEASIBILITY_SENSE` (ignore the objective).

ianfiske: Thank you for catching that and bonus for pointing me to the location in the Clp source!  Just pushed a fix.

@odow I briefly looked add the idea of adding a test, but seems that the tests are owned by MOI -- would that be the place to add such a test?

odow: Just add a new test to the bottom of https://github.com/JuliaOpt/Clp.jl/blob/master/test/MOIWrapper.jl.
Create a model, set the objective sense to -3, and then query `LQOI.get_objective_sense`. Alternatively, just change the `elseif` prior to the `FEASIBILITY_SENSE` into an `else` and remove the error.

ianfiske: Thank @odow -- I just pushed a change with your alternative suggestion.

src/MOIWrapper.jl
```diff
@@ -82,6 +82,19 @@ function replace_inf(x::Real)
...
 
+
+function LQOI.get_objectivesense(model::Optimizer)
+    s = get_obj_sense(model.inner)
+    if s == 1.0
+        return MOI.MIN_SENSE
+    elseif s == -1.0
+        return MOI.MAX_SENSE
+    else
+        return MOI.FEASIBILITY_SENSE
```
blegat: I would add `@assert iszero(s)` before the return just to be sure

ianfiske: Thanks @blegat -- just added that.

mlubin:
Thanks! I'll tag right now.

PR #55, opened 2019-02-27
Unsafe get function

ndinsmore:
This creates get functions which return unsafe arrays and are annotated as such.   It then uses those function where it appears safe.  There should be a slight performance improvement due to fewer memory allocations.

But the chief reason for this improvement is the function ```unsafe_get_constraint_matrix```  which can be used to modify the constraint matrix.

The pull request:
https://github.com/JuliaOpt/Clp.jl/pull/56
Takes advantage of this.

### DIFF
--- a/src/ClpCInterface.jl
+++ b/src/ClpCInterface.jl
@@ -81,6 +81,7 @@ export
     get_vector_starts,
     get_indices,
     get_constraint_matrix,
+    unsafe_get_constraint_matrix,
     get_vector_lengths,
     get_elements,
     get_obj_value,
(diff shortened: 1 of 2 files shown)


mlubin:
> But the chief reason for this improvement is the function `unsafe_get_constraint_matrix` which can be used to modify the constraint matrix.

`Clp_getElements` returns a `const double *`:
https://github.com/coin-or/Clp/blob/3578599f862de64724db82f83dfa81be16a45e02/Clp/src/Clp_C_Interface.h#L228

I interpret from this that it's not intended for this function to be used to modify elements in the constraint matrix.

ndinsmore:
Yes, but it does work and adds a good amount of functionality.

mlubin:
The Clp API supports modifying coefficients now: https://github.com/coin-or/Clp/issues/102. Closing this PR as no longer necessary.

PR #56, opened 2019-02-27
add support for LQOI.change_matrix_coefficient

ndinsmore:
This adds support for ```LQOI.change_matrix_coefficient```. which allows JuMP to use ```set_coefficient``` on constraint variables which previously had been created.  Even with the limitation, it adds a lot of capability to the MOI Clp Interface.

This is a hack that uses the fact the clp_getElement returns a pointer to the actual matrix values.  It is an abuse of the interface.... But it works.

That said I am a bit out of my depth in writing a MOI test that ensures that we can check that this hack continues to work.   

The JuMP test below proves it works
```julia
using Clp
using JuMP
using Test

function coefficient_modification_test(num_vars )
    model = Model(with_optimizer(Clp.Optimizer; LogLevel=0))

    test_vars=@variable(model,test_vars[i=1:num_vars])
    test_bound_var=@variable(model,0<=test_bound_var<=1)

    cons=@constraint(model,test_vars .- (1*test_bound_var) .== 0)
    @objective(model,Max,sum(test_vars))

    optimize!(model)

    if objective_value(model) !=  num_vars
        println("Initial objective was not what was expected :   $(objective_value(model)) != $num_vars")
    else
        println("Initial objective correct")
    end

    for con in cons
        set_coefficient(con,test_bound_var,-2)
    end

    optimize!(model)

    @test objective_value(model) ==  2*num_vars

    if objective_value(model) !=  2*num_vars
        println("After modification objective was not what was expected :   $(objective_value(model)) != $(2*num_vars)")
    else
        println("After modification objective correct")
    end

end
```

### DIFF
--- a/src/ClpCInterface.jl
+++ b/src/ClpCInterface.jl
@@ -81,6 +81,7 @@ export
     get_vector_starts,
     get_indices,
     get_constraint_matrix,
+    unsafe_get_constraint_matrix,
     get_vector_lengths,
     get_elements,
     get_obj_value,
--- a/src/ClpCInterface.jl
+++ b/src/ClpCInterface.jl
@@ -811,46 +811,65 @@ function get_num_elements(model::ClpModel)
 end
 
 # Get column starts in matrix.
-function get_vector_starts(model::ClpModel)
+# Unsafe because array is in memory which is not managed by julia
+function unsafe_get_vector_starts(model::ClpModel)
     _jl__check_model(model)
     vec_starts_p = @clp_ccall getVectorStarts Ptr{CoinBigIndex} (Ptr{Cvoid},) model.p
     num_cols = Int(get_num_cols(model))
-    return copy(unsafe_wrap(Array,vec_starts_p, (num_cols+1,)))
+    return unsafe_wrap(Array,vec_starts_p, (num_cols+1,))
 end
 
+# Get column starts in matrix.
+get_vector_starts(model::ClpModel) = copy(unsafe_get_vector_starts(model))
+
 # Get row indices in matrix.
-function get_indices(model::ClpModel)
+# Unsafe because array is in memory which is not managed by julia
+function unsafe_get_indices(model::ClpModel)
     _jl__check_model(model)
     # getIndices returns an "int*", how do we know it's Int32??
     row_indices_p = @clp_ccall getIndices Ptr{Int32} (Ptr{Cvoid},) model.p
     num_elts = Int(get_num_elements(model))
-    return copy(unsafe_wrap(Array,row_indices_p,(num_elts,)))
+    return unsafe_wrap(Array,row_indices_p,(num_elts,))
 end
 
-function get_constraint_matrix(model::ClpModel)
+# Get row indices in matrix.
+get_indices(model::ClpModel) = copy(unsafe_get_indices(model))
+
+# Get the constraint matrix.
+# Unsafe because array is in memory which is not managed by julia.
+function unsafe_get_constraint_matrix(model::ClpModel)
     _jl__check_model(model)
     @assert CoinBigIndex == Int32
     # SparseMatrixCSC requires same integer type for colptr and rowval
     num_cols = convert(Int, get_num_cols(model))
     num_rows = convert(Int, get_num_rows(model))
-    colptr = get_vector_starts(model) .+ convert(Int32,1)
-    rowval = get_indices(model) .+ convert(CoinBigIndex,1)
-    nzval = get_elements(model)
-
+    colptr = unsafe_get_vector_starts(model) .+ convert(Int32,1)
+    rowval = unsafe_get_indices(model) .+ convert(CoinBigIndex,1)
+    nzval = unsafe_get_elements(model)
     return SparseMatrixCSC{Float64,CoinBigIndex}(num_rows,num_cols,colptr,rowval,nzval)
+end
 
+# Get the constraint matrix.
+function get_constraint_matrix(model::ClpModel)
+    mat = unsafe_get_constraint_matrix(model)
+    return SparseMatrixCSC(mat.m, mat.n, mat.colptr, mat.rowval, copy(mat.nzval))
 end
 
 # Get column vector lengths in matrix.
 @def_get_col_property get_vector_lengths getVectorLengths
 
 # Get element values in matrix.
-function get_elements(model::ClpModel)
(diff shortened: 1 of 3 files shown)


mlubin:
It's nice to have problem modification working for Clp, but "it works on one example" is not a sufficiently convincing argument to justify abusing the API. Hacks like these introduce a lot of technical debt that the maintainers of Clp.jl will have to deal with down the line (e.g., by addressing bug reports when it *doesn't* work). A convincing justification would be one of:
1. A statement from the [Clp mailing list](https://list.coin-or.org/mailman/listinfo/clp) that this hack is ok and safe.
2. An analysis of the Clp source code explaining why this field is safe to manipulate. How does it interact with presolve, for example?
3. A change to the Clp C API to remove the `const`, or an addition to the Clp C API that provides for modifying coefficients.

ndinsmore:
The following is an analysis of the code base:
1.)  First lets look at the function whose output we are abusing:
https://github.com/coin-or/Clp/blob/3578599f862de64724db82f83dfa81be16a45e02/Clp/src/Clp_C_Interface.cpp#L634-L639
There is a little macro magic going on here but the important thing to understand is ```model_ ```and object of class ```ClpSimplex```.  This relation can be seen in:
https://github.com/coin-or/CoinUtils/blob/eb7cd1a0575921a06d69b9f2a8a494dc3f853092/CoinUtils/src/Coin_C_defines.h#L53-L57
From that code it is clear the ```ClpSimplex``` has a method ```matrix()``` which returns the matrix.
2.)  The method ```matrix()``` which is public  is actually in a super class of ```ClpSimplex``` <: ```ClpModel```  :
https://github.com/coin-or/Clp/blob/3578599f862de64724db82f83dfa81be16a45e02/Clp/src/ClpModel.hpp#L768
```matrix_``` is a property of ```ClpModel`` seen here:
https://github.com/coin-or/Clp/blob/3578599f862de64724db82f83dfa81be16a45e02/Clp/src/ClpModel.hpp#L1275-L1276
3.)  Interestingly ```ClpModel``` also has a public method called ```modifyCoefficient```:
https://github.com/coin-or/Clp/blob/3578599f862de64724db82f83dfa81be16a45e02/Clp/src/ClpModel.hpp#L229-L234
which clearly and simply modifies a single matrix coeffient using the ```modifyCoefficient``` of the matrix class.

4.) ```modifyCoefficient``` is very simple if you ignore insertion or deletion of elements:
https://github.com/coin-or/CoinUtils/blob/eb7cd1a0575921a06d69b9f2a8a494dc3f853092/CoinUtils/src/CoinPackedMatrix.cpp#L375-L376
with the changing of the element being as simple as:
https://github.com/coin-or/CoinUtils/blob/eb7cd1a0575921a06d69b9f2a8a494dc3f853092/CoinUtils/src/CoinPackedMatrix.cpp#L395
Importantly there is no flag being set of method called to signify that this matrix has been changed.

5.) Importantly that same ```getElements()``` method we saw in step one is providing direct access to ```elements_``` :
https://github.com/coin-or/CoinUtils/blob/eb7cd1a0575921a06d69b9f2a8a494dc3f853092/CoinUtils/src/CoinPackedMatrix.hpp#L126

6.) So while we may be abusing the interface by ignoring const, we are doing nothing different than the public method ```modifyCoefficient```.

7.) Even though it is clear that the actions we have taken are allowed at the c++ level.   When we look at presolve is becomes clear that the matrix we changed is the only way that constraint information actually gets transmitted in.
https://github.com/coin-or/Clp/blob/8419e63eebec28734f273cd608cb683943305dbc/Clp/src/ClpPresolve.cpp#L103-L122
with ```clpMatrix()``` being no different that ```matrix()``` above.

**Comment by mlubin** [comment]
Ok thanks. That's a good analysis. What about adding something like
```c
void Clp_modifyCoefficient(Clp_Simplex *model, int row, int column, double newElement,
  bool keepZero)
{
  model->model_->modifyCoefficient(row, column, newElement, keepZero);
}
```
to the Clp C API? They should accept pull requests, and @odow is in touch with the COIN-OR folks and can bug them if they take too long to respond.

Benefits:
- Simplifies the Clp.jl (no need to keep a cache of the matrix and add long explanations of why we're abusing the API)
- Works for modifying coefficients that aren't currently nonzero.

Downsides:
- It will take a bit longer to release because we'll need to update to a new version of Clp.

ndinsmore:
I think that my permission would be for both routes actually.  Ideally, in the same discussion with the Clp team, get the okay to directly modify the matrix as well.   

The reason is this whole endeavor for me started when I needed to move my problem to a column generation paradigm.   With that I am doing a bunch of changing coefficients, to the point I would start to worry about the overhead of all those `ccall`s

mlubin:
> With that I am doing a bunch of changing coefficients, to the point I would start to worry about the overhead of all those ccalls

There's a difference between starting to worry and explicitly trading off good API design for performance. I'd want to see a benchmark where the `ccall` overhead of changing coefficients is a significant overhead in the whole solve procedure. I'm a bit doubtful because the `ccall` overhead should be about as or possibly less expensive than looking up the position of the coefficient in the sparse matrix. If you're changing a previously zero coefficient to nonzero then the cost of the `ccall` is certainly negligible compared with memory allocations and the cost of moving lots of elements around.

Why do you need to modify coefficients for column generation, by the way? There's a specialized API for adding columns with coefficients provided: https://github.com/coin-or/Clp/blob/master/Clp/src/Clp_C_Interface.h#L123. If we're going to hack things for column generation we should aim to use that API instead of modifying coefficients.

ndinsmore:
@mlubin changing coefficents for column generation really comes into play when you have hit a "max number of columns" and would like to start replacing columns rather than adding new ones.   

In the MOI documentation it says:
> Column Generation
There is no special interface for column generation. If the solver has a special API for setting coefficients in existing constraints when adding a new variable, it is possible to queue modifications and new variables and then call the solver's API once all of the new coefficients are known.

But I can't see in MOI or LQOI I could add the capability to take advantage of Clp's `add_column` with coefficents.

mlubin:
The Clp C API now has `Clp_modifyCoefficient`. After the next Clp release, we can get the binaries updated and then we'll have a super clean way to implement `LQOI.change_matrix_coefficient`.

mlubin:
The wrapper no longer uses LQOI, so this PR is stale and will need to be revisited if applicable.

PR #60, opened 2019-04-10
remove 0.6-dependent code, invalid 1.0

matbesancon:
Clp was broken on 1.0 due to undefined initialization of arrays, also fixed various things

### DIFF
--- a/src/ClpCInterface.jl
+++ b/src/ClpCInterface.jl
@@ -185,7 +185,7 @@ import Base.pointer
 if isfile(joinpath(dirname(@__FILE__),"..","deps","deps.jl"))
     include("../deps/deps.jl")
 else
-    error("Clp not properly installed. Please run Pkg.build(\"Clp\")")
+    error("Clp not properly installed. Please run import Pkg; Pkg.build(\"Clp\")")
 end
 
 ## Shared library interface setup
@@ -536,14 +536,14 @@ function get_num_rows(model::ClpModel)
     _jl__check_model(model)
     @clp_ccall getNumRows Int32 (Ptr{Cvoid},) model.p
 end
-number_rows = get_num_rows
+const number_rows = get_num_rows
 
 # Number of columns.
 function get_num_cols(model::ClpModel)
     _jl__check_model(model)
     @clp_ccall getNumCols Int32 (Ptr{Cvoid},) model.p
 end
-number_cols = get_num_cols
+const number_cols = get_num_cols
 
 # Get primal tolerance.
 function primal_tolerance(model::ClpModel)
@@ -601,7 +601,7 @@ and add new unit tests.
 # Fill in array with problem name.
 function problem_name(model::ClpModel)
     _jl__check_model(model)
-    problem_name = Array{UInt8}(1000)
+    problem_name = Array{UInt8}(undef, 1000)
     @clp_ccall problemName Cvoid (Ptr{Cvoid}, Int32, Ptr{UInt8}) model.p 1000 problem_name
     return bytestring(problem_name)
 end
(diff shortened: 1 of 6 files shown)


src/ClpCInterface.jl
```diff
@@ -620,13 +620,13 @@ function number_iterations(model::ClpModel)
...
 end
-get_iteration_count = number_iterations
+const get_iteration_count = number_iterations
 
 # Set number of iterations
 function set_number_iterations(model::ClpModel, iters::Integer)
     _jl__check_model(model)
     @clp_ccall setNumberIterations Cvoid (Ptr{Cvoid}, Int32) model.p iters
-    return
+    return nothing
```
mlubin: Use `return` instead of `return nothing`: http://www.juliaopt.org/JuMP.jl/dev/style/#Return-statements-1.

matbesancon: ok sorry I had missed this part of the style guide

matbesancon:
Shouldn't the test be added to `MOIT`.
The simplest way to reproduce it is an unbounded problem. With JuMP:
```julia
using JuMP; import Clp
julia> const m = Model(with_optimizer(Clp.Optimizer, LogLevel = 0))
julia> @variable(m, 0 <= x[1:5])
julia> @objective(m, Max, sum(x))
julia> optimize!(m)
ERROR: MethodError: no method matching Array{Float64,N} where N(::Int64)
Closest candidates are:
  Array{Float64,N} where N(::UndefInitializer, ::Int64) where T at boot.jl:416
  Array{Float64,N} where N(::UndefInitializer, ::Int64, ::Int64) where T at boot.jl:417
  Array{Float64,N} where N(::UndefInitializer, ::Int64, ::Int64, ::Int64) where T at boot.jl:418
  ...
Stacktrace:
 [1] unbounded_ray(::Clp.ClpCInterface.ClpModel) at /home/mbesancon/.julia/packages/Clp/zGJX2/src/ClpCInterface.jl:894

```

odow:
Shouldn't `linear8c` have caught this?

https://github.com/JuliaOpt/MathOptInterface.jl/blob/bf36784e58ddcfcb3751e07283f8842c327f3416/src/Test/contlinear.jl#L996-L1033

matbesancon:
there seems to be only a subset of the tests added to Clp, at least that's how I understand `test/MOI_wrapper.jl`

odow:
Those are the excluded tests.

test/MOIWrapper.jl
```diff
@@ -24,6 +24,7 @@ end
         # linear1 test is disabled due to the following bug:
         # https://projects.coin-or.org/Clp/ticket/84
         "linear1",
+        "linear8c",
```
mlubin: This is the list of excluded tests. We want to run this test.

matbesancon: oh got it

mlubin:
So if linear8c was already running, why didn't it fail? I'm confused.

matbesancon:
So am I, that's why I thought the argument passed was the vector of tests run, not not run

mlubin:
linear8c was running and passing. The problematic code was run only when Clp had no ray to return:
https://github.com/JuliaOpt/Clp.jl/blob/81c2c9ade1197a5b1133a47fdc80778a6a842d9e/src/ClpCInterface.jl#L890-L895

I guess this happens in corner cases like the one you showed above where the model has no constraints. Not sure that corner case is worth a new test in MOIT. I recommend adding a small test in Clp that triggers the `else` branch.

matbesancon:
alright, I finally managed to have a test breaking under master and passing with the fix

PR #63, opened 2019-09-17
DNMY: Update to MOI v0.9

mtanneau:


mtanneau:
[for tracking the issue with Clp] https://github.com/coin-or/Clp/issues/117
Clp's C interface does not allow to change _one_ constraint bound at a time, which would be needed for `MOI.set(model, MOI.ConstraintSet(), s::S)`.

PR #64, opened 2019-10-07
added project toml

matbesancon:
Long overdue Project.toml on Clp, to better track dependencies, etc

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,0 +1,25 @@
+name = "Clp"
+uuid = "5e71592d-7078-450d-8580-c595f51aa13b"
+authors = ["Mathieu Besançon <mathieu.besancon@gmail.com>"]
+repo = "https://github.com/JuliaOpt/Clp.jl.git"
+version = "0.6.3"
+
+[deps]
+BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
+Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
+LinQuadOptInterface = "f8899e07-120b-5979-ab1d-7b97bb9e1a48"
+MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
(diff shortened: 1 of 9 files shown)


matbesancon:
@mtanneau maybe this before #63 ?

mtanneau:
Closes #62.

I had added a `Project.toml`, in my MOI branch ; I'll just rebase once this is merged.

matbesancon:
cc @blegat

mlubin:
Also delete `REQUIRE`?

Project.toml
```diff
@@ -0,0 +1,25 @@
+name = "Clp"
+uuid = "5e71592d-7078-450d-8580-c595f51aa13b"
+authors = ["Mathieu Besançon <mathieu.besancon@gmail.com>"]
```
blegat: I would not include this field for consistency with other solver repo in JuliaOpt.

matbesancon: sorry hadn't seen it was still there, it got generated

Project.toml
```diff
@@ -0,0 +1,25 @@
...
+repo = "https://github.com/JuliaOpt/Clp.jl.git"
+version = "0.6.3"
+
+[deps]
+BinaryProvider = "b99e7846-7c00-51b0-8f62-c81ae34c0232"
+Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
```
blegat: We can remove it as we require Julia v1.0

mlubin:
We should also drop 0.7 from travis and appveyor.

Project.toml
matbesancon: it seems MPB requires Compat for its tests

matbesancon: Compat can be dropped when we don't support MPB

matbesancon:
Removing Compat all together is for another time, since it requires removing MPB

matbesancon:
all tests passing here

blegat:
> Removing Compat all together is for another time, since it requires removing MPB

You can keep `Compat` only for the `test` target

matbesancon:
Just did

Issue #65, opened 2019-10-07
maximum_seconds returns weird results


mtanneau:
Clp returns weird results when querying the time limit _after it has been set_.

The code below
```julia
import Clp
m = Clp.ClpModel()

@show Clp.maximum_seconds(m)  # returns -1.0
Clp.set_maximum_seconds(m, 1.0)
@show Clp.maximum_seconds(m)  # returns a (seemingly) random value
Clp.set_maximum_seconds(m, 0.0)
@show Clp.maximum_seconds(m)  # returns a (seemingly) random value
```
returns seemingly random values after the time limit is set.

Might there be a bug on the C interface's side?

odow:
The issue is that when setting, Clp adds a `+CoinCpuTime()` offset:
https://github.com/coin-or/Clp/blob/29a3d29d94f102e9029eb4be72cde2bfd378d752/Clp/src/ClpModel.cpp#L2793
but it doesn't subtract this when returning.
https://github.com/coin-or/Clp/blob/633bcb884303806a0b083834f702c597f44c01a9/src/ClpModel.hpp#L406

You could open an upstream issue?

mtanneau:
> You could open an upstream issue?

Done. https://github.com/coin-or/Clp/issues/141

PR #66, opened 2019-10-08
Update to MOI v0.9

mtanneau:
As of now, the MOI re-write is partially working.
Still some work to do but I think most heavy-lifting is done :sweat_smile: 

## Existential questions
* What is the role of `Clp.ClpSolve`? So far, I have not integrated that in the MOI wrapper.
  Sounds like I should, but I don't know how it integrates with the rest.
* It seems like Clp negates the constant term in the objective. Passing `-f.constant` to `set_objective_offset` seems to have fixed problems. Not sure whether it's the actual Clp convetion, or just me who did something wrong.

## Things that remain to be implemented
* [x] Currently not supported MOI attributes:
  * [x] `TimeLimitSec`: Clp behaves weird, cf #65 
  [EDIT: not supported until Clp's behaviour is fixed]
  * [x] `SolveTime`: I did not find the corresponding function is Clp
  Possible workaround: use an `@elapsed` when calling the Clp-level solve, and store that value at the wrapper level.
[EDIT: not supported]
  * [x] `DualObjectiveValue`: I'm not sure whether Clp has a corresponding function.
  I thought it would be `Clp.dual_bound`, but that essentially returned + infinity when I tried it
  [EDIT: not supported]
  * [x] `NumberOfThreads`: I don't know whether Clp is always single-threaded (in which case we could return 1 if queried, but prevent the user from setting it), or if this should just not be supported. Did not find a thread-related function in the C interface.
[EDIT: not supported]
* [x] Quadratic objective: Clp supports it but I have not included it (yet)
  [EDIT: left for future work]
* [x] Dupplicate terms in `ScalarAffineFunction` constraints. I do not check that at the wrapper level, rather I'm assuming that the cache automatically gives functions in canonical form
* [x] Print levels: although `MOI.Silent` is set to `false` at the beginning of the tests, and its value is kept intact when `MOI.empty!` is called, Clp still prints during the tests. Help welcome.
* [x] Keep track of parameters/options at the wrapper level. I'm not sure I'm 100% sure of what the convention w.r.t parameters and `empty!`. Should parameters also be reset to their default values?

## Tests that still fail
Non-exhaustive list. See which tests I removed [here](https://github.com/mtanneau/Clp.jl/blob/MOIv0.9/test/MOIWrapper.jl)
* `solve_result_index`: I removed it from the tests, it calls `MOI.get(m, MOI.ObjectiveValue(2))`.
  Not sure how this should be handled here.
* Some tests throw errors while using `VectorAffineFunction` or `VectorOfVariables`. I though these should be handled by the bridge. See for instance error message below
```julia
delete_nonnegative_variables: Test Failed at /home/mathieu/.julia/packages/MathOptInterface/iPJzh/src/Test/UnitTests/variables.jl:114
  Expression: !(MOI.is_valid(model, cv))
Stacktrace:
 [1] delete_nonnegative_variables(::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{Clp.Optimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}, ::MathOptInterface.Test.TestConfig{Float64}) at /home/mathieu/.julia/packages/MathOptInterface/iPJzh/src/Test/UnitTests/variables.jl:114
 [2] macro expansion at /home/mathieu/.julia/packages/MathOptInterface/iPJzh/src/Test/config.jl:47 [inlined]
 [3] macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Test/src/Test.jl:1083 [inlined]
 [4] unittest(::MathOptInterface.Bridges.LazyBridgeOptimizer{MathOptInterface.Utilities.CachingOptimizer{Clp.Optimizer,MathOptInterface.Utilities.UniversalFallback{MathOptInterface.Utilities.Model{Float64}}}}, ::MathOptInterface.Test.TestConfig{Float64}, ::Array{String,1}) at /home/mathieu/.julia/packages/MathOptInterface/iPJzh/src/Test/config.jl:47
```
* `linear8a`: some of the duals appear to be wrong it's likely something on my end.
[EDIT: `linear8a` and `linear12` were excluded from the tests]


odow:
> What is the role of Clp.ClpSolve? So far, I have not integrated that in the MOI wrapper.
Sounds like I should, but I don't know how it integrates with the rest.

Not sure. I'd just follow the MPB wrapper.

> It seems like Clp negates the constant term in the objective. Passing -f.constant to set_objective_offset seems to have fixed problems. Not sure whether it's the actual Clp convetion, or just me who did something wrong.

Is this a `min`/`max` problem?

> Quadratic objective: Clp supports it but I have not included it (yet)

No rush to include.  Can be latest PR. But not sure how well it's tested.

> Dupplicate terms in ScalarAffineFunction constraints. I do not check that at the wrapper level, rather I'm assuming that the cache automatically gives functions in canonical form

You should not assume that the passed function is canonicalized; call `MOI.Utilities.canonical`.

> Print levels: although MOI.Silent is set to false at the beginning of the tests, and its value is kept intact when MOI.empty! is called, Clp still prints during the tests. Help welcome.

> Keep track of parameters/options at the wrapper level. I'm not sure I'm 100% sure of what the convention w.r.t parameters and empty!. Should parameters also be reset to their default values?

I think we've kept the parameters. 

> solve_result_index: I removed it from the tests, it calls MOI.get(m, MOI.ObjectiveValue(2)).

You should throw `MOI.ResultIndexBoundsError` or call [MOI.check_result_index_bounds(model, attr)](https://github.com/JuliaOpt/MathOptInterface.jl/blob/e29f1cedf94be20b7baa52ba4f9ad88e917a8988/src/attributes.jl#L130-L135)

> Some tests throw errors while using VectorAffineFunction or VectorOfVariables. I though these should be handled by the bridge. See for instance error message below

Bug in MOI. Wait for new tag.

> linear8a: some of the duals appear to be wrong it's likely something on my end.

Not sure.

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
+MOI.get(model::Optimizer, ::MOI.TimeLimitSec) = Clp.maximum_seconds(model.inner)
+
+MOI.supports(::Optimizer, ::MOI.NumberOfThreads) = false
+
+
+# =====================
+# 
+# =======================
+MOI.supports_constraint(::Optimizer,
+    ::Type{MOI.ScalarAffineFunction{Float64}}, ::Type{S}
```
blegat: You can do `Type{<:SCALAR_SETS}` directly

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
-        if lb > -Inf
-            rhs = lb
-        elseif ub < Inf
-            rhs = ub
-        else
-            error("Either row_lower or row_upper must be of abs less than 1e20")
+MOI.supports(::Optimizer,
+    ::MOI.ObjectiveFunction{<:Union{
+        MOI.ScalarAffineFunction{Float64},
+        MOI.SingleVariable
```
blegat: You don't have to support `MOI.SingleVariable`, there is a bridge for transforming it to `ScalarAffineFunction`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
+)
+    # Check that all constraints and objectives
+    for (F, S) in MOI.get(src, MOI.ListOfConstraints())
+        MOI.supports_constraint(dest, F, S) || throw(MOI.UnsupportedConstraint{F, S}(
+            "Clp.Optimizer does not support constraints of type $F-in-$S."
+        ))
+    end
+    fobj_type = MOI.get(src, MOI.ObjectiveFunctionType())
+    if ! MOI.supports(dest, MOI.ObjectiveFunction{fobj_type}())
+        error("Unsupported objective $(fobj_type).")
```
blegat: Throw `MOI.UnsupportedAttribute`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
 
-function LQOI.set_linear_objective!(instance::Optimizer, cols::Vector{Int}, coefs::Vector{Float64})
-    objective_coefficients = zeros(Float64, get_num_cols(instance.inner))
-    for (col, coef) in zip(cols, coefs)
-        objective_coefficients[col] += coef
+    # Create linear constraints
+    nrows = 0  # number of rows in the problem
+    for (F, S) in MOI.get(src, MOI.ListOfConstraints())
+        # Check if constraint is a linear constraint
+        (F == MOI.ScalarAffineFunction{Float64}) || continue
```
blegat: You should throw `MOI.UnsupportedConstraint` if it's not supported, here it will just be silently ignored

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
+
+        for index in MOI.get(src, MOI.ListOfConstraintIndices{F, S}())
+            # Retrieve problem data
+            # TODO: ensure that `(f, s)` are in canonical form
+            f = MOI.get(src, MOI.ConstraintFunction(), index)
```
blegat: Call `MOIU.canonical`, bridges may create duplicates

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
-        elseif haskey(solveoptionmap, name)
-            solveoptionmap[name](solveroptions,value)
-        else
-            error("Unrecognized option: $name")
+    fobj = MOI.get(src, MOI.ObjectiveFunction{fobj_type}())
+    obj_coeffs = zeros(length(x_src))
+    obj_offset = 0.0
+
+    if fobj_type == MOI.SingleVariable
+        x = mapping.varmap[fobj.variable]
```
blegat: You should not support it as Clp does not exploit its structure, it's just treating it as an affine objective

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
+    elseif fobj_type == MOI.ScalarAffineFunction{Float64}
+        # Record objective coeffs
+        for term in fobj.terms
+            x = mapping.varmap[term.variable_index]
+            # Here we do += instead of = to accumulate dupplicates
+            obj_coeffs[x.value] += term.coefficient
         end
+        obj_offset = fobj.constant
+    else
+        error("Unsupported objective type: $(fobj_type).")
```
blegat: Throw `MOI.UnsupportedAttribute`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
         return MOI.FEASIBLE_POINT
     else
         return MOI.UNKNOWN_RESULT_STATUS
     end
 end
 
-function LQOI.get_dual_status(instance::Optimizer)
-    if is_proven_primal_infeasible(instance.inner)
+function MOI.get(model::Optimizer, ::MOI.DualStatus)
+    if Clp.is_proven_primal_infeasible(model.inner)
```
blegat: Call `MOI.check_result_index_bounds(model, attr)`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
         return MOI.FEASIBLE_POINT
     else
         return MOI.UNKNOWN_RESULT_STATUS
     end
 end
 
-function LQOI.get_number_variables(instance::Optimizer)
-    return get_num_cols(instance.inner)
+function MOI.get(model::Optimizer, ::MOI.VariablePrimal, x::MOI.VariableIndex)
+    # We assume the variable index is valid
```
blegat: Call `MOI.check_result_index_bounds(model, attr)`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
+    sol = Clp.get_col_solution(model.inner)
+    return [sol[x.value] for x in xs]
 end
 
-function LQOI.delete_variables!(instance::Optimizer, start_col::Int, end_col::Int)
-    delete_columns(instance.inner, [Cint(i-1) for i in start_col:end_col])
+function MOI.get(model::Optimizer, ::MOI.ConstraintPrimal,
+    c::MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64}, S}
+) where{S <:SCALAR_SETS}
+    # TODO: Check that constraint index is valid
```
blegat: Call `MOI.check_result_index_bounds(model, attr)`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
-    row = instance[ci]
-    stat = STATMAP[get_row_status(instance.inner, row)]
-    # Single sided constraints should not specify `NONBASIC_AT_X` but only `NONBASIC`.
-    if stat == MOI.NONBASIC_AT_LOWER || stat == MOI.NONBASIC_AT_UPPER
-        return MOI.NONBASIC
-    end
-    return stat
+MOI.get(model::Optimizer, ::MOI.ConstraintPrimal,
+    c::MOI.ConstraintIndex{MOI.SingleVariable, S}
+) where{S <: SCALAR_SETS} = MOI.get(model, MOI.VariablePrimal(), MOI.VariableIndex(c.value))
```
blegat: Call `MOI.check_result_index_bounds(model, attr)`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
+
+function MOI.get(model::Optimizer, ::MOI.ConstraintDual,
+    c::MOI.ConstraintIndex{MOI.ScalarAffineFunction{Float64}, S}
+) where{S <: SCALAR_SETS}
+    dsol = Clp.get_row_price(model.inner)
```
blegat: Call `MOI.check_result_index_bounds(model, attr)`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
-        vi::MOI.ConstraintIndex{MOI.SingleVariable, S}) where S <: Union{LQOI.LE, LQOI.GE, LQOI.EQ, LQOI.IV}
-    col = instance.variable_mapping[instance[vi]]
-    stat = STATMAP[get_column_status(instance.inner, col)]
-    # If, e.g., a column is `NONBASIC_AT_LOWER` then the ≤ constraint is `BASIC`.
-    if (S <: LQOI.LE && stat == MOI.NONBASIC_AT_LOWER) || (S <: LQOI.GE && stat == MOI.NONBASIC_AT_UPPER)
-        return MOI.BASIC
+function MOI.get(model::Optimizer, ::MOI.ConstraintDual,
+    c::MOI.ConstraintIndex{MOI.SingleVariable, MOI.LessThan{Float64}}   
+)
+    rc = Clp.get_reduced_cost(model.inner)[c.value]
```
blegat: Call `MOI.check_result_index_bounds(model, attr)`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
     end
-    # Single sided constraints should not specify `NONBASIC_AT_X` but only `NONBASIC`.
-    if !(S <: LQOI.IV) && (stat == MOI.NONBASIC_AT_LOWER || stat == MOI.NONBASIC_AT_UPPER)
-        return MOI.NONBASIC
+end
+
+function MOI.get(model::Optimizer, ::MOI.ConstraintDual,
+    c::MOI.ConstraintIndex{MOI.SingleVariable, MOI.GreaterThan{Float64}}   
+)
+    rc = Clp.get_reduced_cost(model.inner)[c.value]
```
blegat: Call `MOI.check_result_index_bounds(model, attr)`

Project.toml
```diff
@@ -6,15 +6,14 @@ version = "0.6.3"
...
-LinQuadOptInterface = "f8899e07-120b-5979-ab1d-7b97bb9e1a48"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MathOptInterface = "b8f27783-ece8-5eb3-8dc8-9495eed66fee"
 MathProgBase = "fdba3010-5040-5b88-9595-932c9decdf73"
 SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 
 [compat]
 BinaryProvider = "0.5"
-MathOptInterface = "0.8"
+MathOptInterface = "0.9"
```
blegat: Require `v0.9.4` for `MOI.check_result_index_bounds`

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
+#   empty functions
+# ====================
+function MOI.is_empty(model::Optimizer)
+    # A problem is empty if it has no variable and no linear constraint
+    return (Clp.get_num_rows(model.inner) == 0) && (Clp.get_num_cols(model.inner) == 0)
 end
 
+function MOI.empty!(model::Optimizer)
+    # TODO: keep track of parameters
+    model.inner = Clp.ClpModel()
```
frapac: Shouldn't we free the previous model there (unless we would like to fully rely on the garbage collector)?

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
-function append_row(instance::Optimizer, row::Int, lower::Float64,
-                    upper::Float64, rows::Vector{Int}, cols::Vector{Int},
-                    coefs::Vector{Float64})
-    indices = if row == length(rows)
-        rows[row]:length(cols)
+function MOI.set(model::Optimizer, param::MOI.RawParameter, value)
+    # There is now `set_param`-like function in Clp's C interface.
+    # So we go through the list of individual parameters
+    if param.name == :LogLevel
+        Clp.set_log_level(model.inner, value)
```
frapac: Update the attribute `silent` in `model`?

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
     end
+    return nothing
 end
 
-function LQOI.get_linear_constraint(instance::Optimizer, row::Int)::Tuple{Vector{Int}, Vector{Float64}}
-    A = get_constraint_matrix(instance.inner)
-    A_row = A[row,:]
-    return Array{Int}(A_row.nzind), A_row.nzval
+function MOI.get(model::Optimizer, ::MOI.Silent)
+    return Clp.log_level(model.inner) == 0
```
frapac: Idem, should we set `silent` to `false` there?

src/MOIWrapper.jl
```diff
@@ -1,408 +1,435 @@
...
+        # Possible bound combinations:
+        #=
+            * No bound (default case)
+            * Interval
+            * EqualTo
+            * LessThan
+            * GreaterThan
+            * LessThan && GreaterThan
+        =#
+        lb, ub = -Inf, Inf  # Default case: free variable
```
frapac: Are we sure that `Inf` is correctly specified to Clp, and converted automatically to `DBL_MAX`?

src/MOIWrapper.jl
mtanneau: What does `MOI.check_result_index_bounds(model, attr)` do exactly? Is it to check that a result does exist before returning one?

src/MOIWrapper.jl
mtanneau: Does the function exist in the julia wrapper? Or do we need a direct `ccall`?

src/MOIWrapper.jl
mtanneau: It does.
```julia
import Clp

m = Clp.ClpModel()
Clp.add_column(m, Cint(0), Int32[], Float64[], 0.0, Inf, 1.0)
Clp.get_col_upper(m)
```
returns
```julia
1-element Array{Float64,1}:
 1.7976931348623157e308
```

src/MOIWrapper.jl
frapac: The following function should do the job:
https://github.com/JuliaOpt/Clp.jl/blob/master/src/ClpCInterface.jl#L230-L237

mtanneau:
> > What is the role of Clp.ClpSolve? So far, I have not integrated that in the MOI wrapper.
> > Sounds like I should, but I don't know how it integrates with the rest.
> 
> Not sure. I'd just follow the MPB wrapper.

This is next on the TODO list.

> 
> > It seems like Clp negates the constant term in the objective. Passing -f.constant to set_objective_offset seems to have fixed problems. Not sure whether it's the actual Clp convetion, or just me who did something wrong.
> 
> Is this a `min`/`max` problem?

Doesn't appear so. I tried with baby examples, and the objective offset seems to be always negated.
All related tests are passing, so I guess it's just Clp's convention.

> 
> > Quadratic objective: Clp supports it but I have not included it (yet)
> 
> No rush to include. Can be latest PR. But not sure how well it's tested.
> 

It will be left for future work then

> > Keep track of parameters/options at the wrapper level. I'm not sure I'm 100% sure of what the convention w.r.t parameters and empty!. Should parameters also be reset to their default values?
> 
> I think we've kept the parameters.
> 

That will be done together with the `ClpSolve` part. Coming soon.

> > linear8a: some of the duals appear to be wrong it's likely something on my end.
> 
> Not sure.

These are related to infeasibility rays taking wrong values. Also causes failures on `linear12`.

mtanneau:
Solver parameters / solve option are now tracked automatically, and they are not discarded during an `MOI.empty!`.
The only (raw) parameter that is currently not supported is `maximumIterations`, see #67 and #69.
Once a fix is implemented, it's just one line to un-comment.

Currently in the pipeline: implement these `check_result_index_bounds` where needed.

The last two things left to do are 1) whether `MOI.DualObjectiveValue` should be supported or not, and 2) find out why tests `linear12` and `linear8a` are failing. It has something to do with querying the dual certificates for an infeasible problem.

blegat:
> It has something to do with querying the dual certificates for an infeasible problem.

Clp is not good with that, see https://github.com/JuliaOpt/Clp.jl/pull/18/ and https://github.com/JuliaOpt/Clp.jl/pull/20/. I would not be surprised if it's a Clp bug rather than a bug in the MOI wrapper

src/MOIWrapper.jl
blegat: Yes, it checks with the value of `ResultCount`

mtanneau: FYI, `MOI.check_result_index_bounds` will throw an error when querying `DualStatus(2)`, while the test [`solve_result_index`](https://github.com/JuliaOpt/MathOptInterface.jl/blob/ee77d85c69b8d32b8455376b40b0698e52312341/src/Test/UnitTests/solve.jl#L141) expects an `MOI.NO_SOLUTION`.

I've implemented things so as to pass tests in `solve_result_index`; I did find it a bit weird that `MOI.get(model, MOI.PrimalStatus(2))` is expected to not return an error, but  `MOI.get(model, MOI.VariablePrimal(2), x)` is expected to. Won't prevent me from sleeping though.

blegat: See https://github.com/JuliaOpt/MathOptInterface.jl/pull/912 for the relevant discussion

mtanneau:
I removed the tests that query infeasibility certificates, and de-activated `DualObjectiveValue` in the test config.

The latter needs MOI0.9.6 to work. Once the new version is tagged all tests should pass.

If that's satisfactory then the PR can be merged then.

Issue #67, opened 2019-10-13
ccall to maximumIterations fails


mtanneau:
Calling `Clp.maximum_iterations` somehow ends up in a `ccall` error:
```julia
import Clp

m = Clp.ClpModel()
Clp.set_maximum_iterations(m, 10)
Clp.maximum_iterations(m)  # throws error
```
throws
```julia
ERROR: ccall: could not find function Clp_maximumIterations in library /mnt/c/Users/mathi/GitHub/Clp.jl/deps/usr/lib/libClp.so
Stacktrace:
 [1] maximum_iterations(::Clp.ClpCInterface.ClpModel) at /mnt/c/Users/mathi/GitHub/Clp.jl/src/ClpCInterface.jl:197
 [2] top-level scope at none:0
```

The function is defined in [Clp's C interface](https://github.com/coin-or/Clp/blob/85f522ee1e208d5b7332c1e40c3af56774a49d20/Clp/src/Clp_C_Interface.cpp#L497), so I don't know where the problem comes from.

odow:
It looks like they called it `maximumIterations` instead of `Clp_maximumIterations`.

mtanneau:
To keep track: https://github.com/coin-or/Clp/issues/119

PR #68, opened 2019-10-14
Fix ccall to maximumIterations

mtanneau:
See #67 

> It looks like they called it `maximumIterations` instead of `Clp_maximumIterations`.

The proposed fix replaces the `@clp_ccall` with a regular `ccall` to the correct C function.

```julia
import Clp

m = Clp.ClpModel()
Clp.maximum_iterations(m)  # no error anymore, should return typemax(Int32)
```


mtanneau:
whoops, Wrong branch history

PR #74, opened 2020-02-19
Fixes parameters mapping for new JuMP string approach

bhalonen:
I may be fairly clueless on what should change, but this made my error go away.


odow:
Close attempt! I pushed an alternative fix in #75 that won't break existing code.

PR #75, opened 2020-02-19
Deal with String RawParameters

odow:
Update for JuMP 0.21.

Closes #74


src/MOIWrapper.jl
```diff
@@ -189,7 +190,7 @@ function MOI.copy_to(
     # Empty dest
     MOI.empty!(dest)
 
-    # Maps the indices of the 
+    # Maps the indices of the
```
mlubin: Delete this incomplete comment?

PR #77, opened 2020-04-03
Add Clp_jll

### LINKED ISSUE #70: Building Clp on Julia 1.3.0-rc3.0
i'm having hard times to build Clp package on julia 1.3.0, i added the package successfully, but when i want to build it, it takes forever without any result .
It's showing me only this :

Building Clp → `~/.julia/packages/Clp/IBQzB/deps/build.log`
i'm on MacOS high sierra

### LINKED ISSUE #73: Library not loaded : @rpath/libgfortran.4.dylib
I write a Julia code with [`Games.jl`](https://github.com/QuantEcon/Games.jl),
```
pd_payoff = [9.0 1.0; 10.0 3.0]
A, B = Player(pd_payoff), Player(pd_payoff)
pd = NormalFormGame((A, B))
rpd = RepeatedGame(pd, 0.75)
hp_pts = outerapproximation(rpd; nH=64, maxiter=500, tol=1e-8, verbose=true, nskipprint=10)
```
Then, I have a following error.
```
error compiling #outerapproximation#62: error compiling Optimizer#1: could not load library "/Users/myname/.julia/packages/Clp/4DejL/deps/usr/lib/libClp.1.13.11.dylib"

dlopen(/Users/myname/.julia/packages/Clp/4DejL/deps/usr/lib/libClp.1.13.11.dylib, 1): Library not loaded: @rpath/libgfortran.4.dylib   
   Referenced from: /Users/myname/.julia/packages/Clp/4DejL/deps/usr/lib/libClp.1.13.11.dylib 
   Reason: image not found
```
My `/Users/myname/.julia/packages/Clp/4DejL/deps/usr/lib/` directory is like this:

```
MyMacBook-Pro-2:~ myname$ ls ~/.julia/packages/Clp/4DejL/deps/usr/lib/
   libClp.1.13.11.dylib     libClpSolver.1.13.11.dylib  libCoinUtils.la         libOsiClp.dylib         libcoinlapack.la
   libClp.1.dylib           libClpSolver.1.dylib        libOsi.la               libOsiClp.la            libcoinmetis.la
   libClp.dylib             libClpSolver.dylib          libOsiClp.1.13.11.dylib libOsiCommonTests.la    libcoinmumps.la
   libClp.la                libClpSolver.la             libOsiClp.1.dylib       libcoinblas.la          pkgconfig
```
Does anyone have an idea to resolve this Library not found error?

This issue comes from [here](https://stackoverflow.com/questions/60183260/library-not-loaded-rpath-libgfortran-4-dylib?noredirect=1#comment106448389_60183260). Thank you @giordano !

odow:
- [x] Needs https://github.com/JuliaRegistries/General/pull/12117

But can confirm that I have passing locally!

x-ref: https://github.com/JuliaPackaging/Yggdrasil/issues/509

Closes #73
Closes #70


giordano:
Note that this should fix #73

ViralBShah:
Given how sensitive the solvers are to having the right versions of everything, I think we should tightly bind versions for all the jll dependencies. The reason being that we will want to start upgrading the dependencies in the artifacts, but not yet want them to be pulled by Clp.jl and GLPK.jl.

ViralBShah:
I assume that even if Clp 1.17 is not compatible with 1.16, we should not be getting the double free kind of malloc related errors in Julia.

odow:
> we should not be getting the double free kind of malloc related errors in Julia.

Agreed.

On the plus side, Cbc is green: https://github.com/JuliaOpt/Cbc.jl/pull/133. Which suggests that it might be an API issue regarding how we're calling Clp.

I couldn't see any obvious offenders if you scan `Clp_C_Interface.cpp` and `Clp_C_Interface.h`:
https://github.com/coin-or/Clp/compare/releases%2F1.16.11...releases%2F1.17.5?diff=split

ViralBShah:
Right, but cbc was already using Clp 1.17 even before. I will take a look at the clp interface.

ViralBShah:
I see `CoinBigIndex` being used. Do you know if this is set to 64-bit on 64-bit architectures? It is unclear if this affects the API

I see this comment in `Clp_C_Interface.h`:
```
/* accidentally used a bool for Clp_modifyCoefficient, so need to include stdbool.h
 * Clp_modifyCoefficient signature will change to use int with Clp 1.18
```

odow:
Yes, I saw the `int`->`CoinBigIndex` changes. It's hard coded to `const CoinBigIndex = Int32` on the Julia wrapper. I think that is the default unless you compile with a different flag. Presumably, if it was actually `Int64` Mac would also fail? I can try and see what happens.
https://github.com/JuliaOpt/Clp.jl/blob/c6121a52c6c980afa231d88944b71d90884494c0/src/ClpCInterface.jl#L210-L213

For now, we don't wrap `Clp_modifyCoefficient`, so that isn't a problem for us.

ViralBShah:
Reading through `CoinUtils`, it seems like `CoinBigIndex` should be `Int32` since I don't see any indication of setting a different value.

ViralBShah:
Do you think that using our own ASL instead of the one Coin-OR wants to use could be an issue here?

mtanneau:
I tried running the tests locally; I can pin down the failure to occurring withing a call to `Clp_deleteModel`.

However, it does not reliably occur during a particular testset.
Initially, it occurred during `MOIT.solve_constant_obj`. I excluded it from the tests, and the same error occurred, simply on another testset.
I was also not able to reproduce the bug by running these testsets individually.

odow:
@mtanneau I was zeroing in on `deleteModel` as well. Can you try this diff:
```Julia
diff --git a/src/MOIWrapper.jl b/src/MOIWrapper.jl
index e3be6ba..8c00a0f 100644
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -70,22 +70,13 @@ end
 
 function MOI.empty!(model::Optimizer)
     old_model = model.inner
-
-    # Create new Clp object
     model.inner = Clp.ClpModel()
-
     # Copy parameters from old model into new model
-    for (option, (getter, setter)) in CLP_OPTION_MAP
-        value = getter(old_model)
-        setter(model.inner, value)
+    for (_, (getter, setter)) in CLP_OPTION_MAP
+        setter(model.inner, getter(old_model))
     end
-
     model.optimize_called = false
-
-    # Free old Clp object
-    Clp.ClpCInterface.delete_model(old_model)
-
-    return nothing
+    return
 end
 ```

**Comment by mtanneau** [comment]
Still happening, rarely in the same place. Sometimes a dozen testsets can run, sometimes only a few.

I tried this diff
```
diff --git a/src/ClpCInterface.jl b/src/ClpCInterface.jl
index 6ca604b..82bb7ca 100644
--- a/src/ClpCInterface.jl
+++ b/src/ClpCInterface.jl
@@ -225,17 +225,21 @@ mutable struct ClpModel
     function ClpModel()
         p = @clp_ccall newModel Ptr{Cvoid} ()
         prob = new(p)
-        finalizer(delete_model, prob)
+        # finalizer(delete_model, prob)
         return prob
     end
 end
```
since `Clp_deleteModel` kept being called, even though I had removed the explicit call to `delete_model` in `MOI.empty!`.

No real change, but I'm wondering whether the GC might be involved in all this.

**Comment by mtanneau** [comment]
(also I don't know what the line I commented out actually does, but I'm guessing it's GC-related)

**Comment by odow** [comment]
This is suspicious: https://travis-ci.org/github/JuliaOpt/Clp.jl/jobs/671755091#L431-L434
```Julia
7f394344d000-7f39436b0000 r-xp 00000000 08:01 2055349                    /home/travis/.julia/artifacts/9cf32528460420740b8ba41eb7a2833b8975b4c2/lib/libClp.so.1.14.5
7f39436b0000-7f39438af000 ---p 00263000 08:01 2055349                    /home/travis/.julia/artifacts/9cf32528460420740b8ba41eb7a2833b8975b4c2/lib/libClp.so.1.14.5
7f39438af000-7f39438b7000 rw-p 00262000 08:01 2055349                    /home/travis/.julia/artifacts/9cf32528460420740b8ba41eb7a2833b8975b4c2/lib/libClp.so.1.14.5
7f39438b7000-7f3943922000 rw-p 002ab000 08:01 2055349                    /home/travis/.julia/artifacts/9cf32528460420740b8ba41eb7a2833b8975b4c2/lib/libClp.so.1.14.5
```
Looks like it built `1.14.5` and not `1.17.5`?

ViralBShah:
In one of the intermediate builds of Clp_jll I had a bad libClp version accidentally. But that should not be happening anymore.

odow:
Well, this seems to be the culprit:

```Julia
Oscars-MBP:artifacts oscar$ ls */lib/libClp*
ls: */lib/libClp*: No such file or directory
Oscars-MBP:artifacts oscar$ ~/julia1.3 --project=/tmp/clp_jll
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.3.1 (2019-12-30)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(clp_jll) pkg> add Clp_jll
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
 Resolving package versions...
  Updating `/private/tmp/clp_jll/Project.toml`
  [06985876] + Clp_jll v1.17.5+2
  Updating `/private/tmp/clp_jll/Manifest.toml`
  [ae81ac8f] + ASL_jll v0.1.1+2
  [06985876] + Clp_jll v1.17.5+2
  [be027038] + CoinUtils_jll v2.11.4+1
  [e66e0078] + CompilerSupportLibraries_jll v0.3.3+0
  [d00139f3] + METIS_jll v4.0.3+0
  [656ef2d0] + OpenBLAS32_jll v0.3.9+1
  [7da25872] + Osi_jll v0.108.6+1
  [2a0f44e3] + Base64 
  [ade2ca70] + Dates 
  [8ba89e20] + Distributed 
  [b77e0a4c] + InteractiveUtils 
  [76f85450] + LibGit2 
  [8f399da3] + Libdl 
  [56ddb016] + Logging 
  [d6f4376e] + Markdown 
  [44cfe95a] + Pkg 
  [de0858da] + Printf 
  [3fa0cd96] + REPL 
  [9a3f8284] + Random 
  [ea8e919c] + SHA 
  [9e88b42a] + Serialization 
  [6462fe0b] + Sockets 
  [8dfed614] + Test 
  [cf7118a7] + UUIDs 
  [4ec0a83e] + Unicode 

(clp_jll) pkg> st Clp_jll
    Status `/private/tmp/clp_jll/Project.toml`
  [06985876] Clp_jll v1.17.5+2

julia> exit()
Oscars-MBP:artifacts oscar$ ls */lib/libClp*
396708ad058ae6c4f6d3945282a357e4cf8c8d63/lib/libClp.1.14.5.dylib
396708ad058ae6c4f6d3945282a357e4cf8c8d63/lib/libClp.1.dylib
396708ad058ae6c4f6d3945282a357e4cf8c8d63/lib/libClp.dylib
396708ad058ae6c4f6d3945282a357e4cf8c8d63/lib/libClp.la
396708ad058ae6c4f6d3945282a357e4cf8c8d63/lib/libClpSolver.1.14.5.dylib
396708ad058ae6c4f6d3945282a357e4cf8c8d63/lib/libClpSolver.1.dylib
396708ad058ae6c4f6d3945282a357e4cf8c8d63/lib/libClpSolver.dylib
396708ad058ae6c4f6d3945282a357e4cf8c8d63/lib/libClpSolver.la
Oscars-MBP:artifacts oscar$ 
```

ViralBShah:
I am using this commit: https://github.com/coin-or/Clp/commit/29a3d29d94f102e9029eb4be72cde2bfd378d752

Isn't that the right one?

odow:
Certainly looks like the right commit. Any explanation for the weird filename versioning?

ViralBShah:
@giordano Is this picking up some older versions somehow from some cache?

odow:
For reference, `libClp` `1.13.11` and `libCbc` `3.10.5` do not exist. So something is weird somewhere.

ViralBShah:
Right those version numbers are weird.

ViralBShah:
FWIW, I checked the installed header files, and those seem to be as expected.

ViralBShah:
> (also I don't know what the line I commented out actually does, but I'm guessing it's GC-related)

Perhaps you know this already. It is just registering a finalizer function, which will be called when a `ClpModel` object is GC'ed by Julia.  This is required to free memory that is allocated inside Clp, which is not visible to Julia's GC.

src/MOIWrapper.jl
```diff
@@ -70,22 +70,13 @@ end
...
-        value = getter(old_model)
-        setter(model.inner, value)
+    for (_, (getter, setter)) in CLP_OPTION_MAP
+        setter(model.inner, getter(old_model))
     end
-
     model.optimize_called = false
-
-    # Free old Clp object
-    Clp.ClpCInterface.delete_model(old_model)
```
ViralBShah, line 86: Yes this could possibly end up calling the finalizer twice.

odow: So this shouldn't have been there, but shouldn't have been the issue because of
https://github.com/JuliaOpt/Clp.jl/blob/c6121a52c6c980afa231d88944b71d90884494c0/src/ClpCInterface.jl#L230-L237
There was plausibly a weird race condition where `delete_model` got called, the function exited, the GC got run, and then `old_model` was finalized, but that doesn't seem to be the issue here.

src/ClpCInterface.jl
```diff
@@ -245,12 +246,14 @@ mutable struct ClpSolve
...
         p = @clpsolve_ccall new Ptr{Cvoid} ()
         prob = new(p)
+        println("Creating ClpSolve: $(p)")
         finalizer(delete_solve, prob)
         return prob
     end
 end
 
 function delete_solve(solve::ClpSolve)
+    println("Finalizing ClpSolve: $(solve.p)")
```
mtanneau, line 256: The docs of `finalizer` says
```julia
finalizer(f, x)

  Register a function f(x) to be called when there are no program-accessible references to x, and return x. The type of x must be a mutable struct, otherwise the behavior of this function is unpredictable.

  f must not cause a task switch, which excludes most I/O operations such as println. @schedule println("message") or ccall(:jl_, Void, (Any,), "message") may be helpful for debugging purposes.
```
Might the `println` risk creating further errors?

src/MOIWrapper.jl
ViralBShah: Yes, that was what I was thinking. CI environments will force GC to be more aggressive due to lesser memory overall. Ideally, we should just let GC clear the object and the finalizer do its job and never free it through other paths. Does that help get rid of the crashes?

src/ClpCInterface.jl
ViralBShah: No, that by itself is not the issue. If there's a race that leads to a double free we need to fix it.

mtanneau:
Looks like the problem arises just after `MOIT.solve_unbounded_model` is run. Removing this testset removed the error (at least the couple of times I tried)
So likely something to do with solving unbounded problems

mtanneau:
By running just `MOIT.solve_unbounded_model` locally, I manage to replicate the bug several times, albeit with sometimes different error messages. For instance:
```julia
free(): invalid next size (fast)
free(): invalid next size (normal)
signal (11): Segmentation fault
```
and sometimes the bug doesn't even appear...

odow:
Okay. I officially give up. It's impossible to debug with without being able to reproduce locally. 

The pointer doesn't appear to be a Julia-visible pointer (e.g., `ClpModel`, `ClpSolve`, or the infeaisibility/unbounded rays).

odow:
If it is `solve_unbounded_model`, that's almost more worrying, because it increases the likelihood that it's an internal Clp bug: https://github.com/JuliaOpt/MathOptInterface.jl/blob/b3f6c71dbbeff73f4664a68a2b2db2b112ce617f/src/Test/UnitTests/solve.jl#L83-L97

It's also weird that this is a Windows/Linux issue, but not Mac. 

It's be nice to trace the reason for the mis-numbered versions. That might give us some insight.

ViralBShah:
I noticed that older clp releases also had the strange version numbers. I will try it with RR tomorrow. We can also try different releases.

Should we ask the Clp folks if they can take a quick look?

ViralBShah:
The tests also immediately crashed in the same place the first time I tried on Linux.

odow:
Progress!!! Looks like it might be an internal Clp bug when calling `status` on an ubounded problem.

https://travis-ci.org/github/JuliaOpt/Clp.jl/jobs/672092738#L865-L870

ViralBShah:
Would it be useful to have login to the linux machine where I see the crash for debugging purposes?

odow:
> Would it be useful to have login to the linux machine where I see the crash for debugging purposes?

💯 This debugging remotely via Travis is getting old... You should have my gmail.

ViralBShah:
[homebrew-coinor](https://github.com/coin-or-tools/homebrew-coinor/blob/master/clp.rb) is using CLP 1.17.3. I can quickly provide that and use versions that are consistent with what they have released. Is it worth trying?

odow:
Sure. Let's try 1.17.3.

odow:
Okay. What a mission.

Weird corruption bug somewhere inside Clp's solve routines under the following conditions:
- model is unbounded
- all variables have a lower bound of `-Inf`.

```Julia
import Clp_jll: libClp

function trigger()
    model = ccall(("Clp_newModel", libClp), Ptr{Cvoid}, ())
    ccall(
        ("Clp_addColumns", libClp),
        Cvoid,
        (Ptr{Cvoid}, Cint, Ptr{Float64}, Ptr{Float64}, Ptr{Float64}, Ptr{Cint}, Ptr{Cint}, Ptr{Float64}),
        model, Cint(1), Float64[-Inf], C_NULL, Float64[1.0], C_NULL, C_NULL, C_NULL
    )
    ccall(("Clp_initialSolve", libClp), Cint, (Ptr{Cvoid},), model)
    ccall(("Clp_deleteModel", libClp), Cvoid, (Ptr{Cvoid},), model)
    return
end

trigger()
```

```
julia> import Clp_jll: libClp

julia> function trigger()
           model = ccall(("Clp_newModel", libClp), Ptr{Cvoid}, ())
           ccall(
               ("Clp_addColumns", libClp),
               Cvoid,
               (Ptr{Cvoid}, Cint, Ptr{Float64}, Ptr{Float64}, Ptr{Float64}, Ptr{Cint}, Ptr{Cint}, Ptr{Float64}),
               model, Cint(1), Float64[-Inf], C_NULL, Float64[1.0], C_NULL, C_NULL, C_NULL
           )
           ccall(("Clp_initialSolve", libClp), Cint, (Ptr{Cvoid},), model)
           ccall(("Clp_deleteModel", libClp), Cvoid, (Ptr{Cvoid},), model)
           return
       end
trigger (generic function with 1 method)

julia> trigger()
Clp3002W Empty problem - 0 rows, 1 columns and 0 elements
Clp0002I Dual infeasible - objective value 0
Clp0032I DualInfeasible objective 0 - 0 iterations time 0.002
free(): invalid pointer

signal (6): Aborted
in expression starting at REPL[33]:1
__libc_signal_restore_set at /build/glibc-OTsEL5/glibc-2.27/signal/../sysdeps/unix/sysv/linux/nptl-signals.h:80 [inlined]
raise at /build/glibc-OTsEL5/glibc-2.27/signal/../sysdeps/unix/sysv/linux/raise.c:48
abort at /build/glibc-OTsEL5/glibc-2.27/stdlib/abort.c:79
__libc_message at /build/glibc-OTsEL5/glibc-2.27/libio/../sysdeps/posix/libc_fatal.c:181
malloc_printerr at /build/glibc-OTsEL5/glibc-2.27/malloc/malloc.c:5350
_int_free at /build/glibc-OTsEL5/glibc-2.27/malloc/malloc.c:4157 [inlined]
__libc_free at /build/glibc-OTsEL5/glibc-2.27/malloc/malloc.c:3124
_ZN8ClpModel12gutsOfDeleteEi at /home/odow/.julia/artifacts/9cf32528460420740b8ba41eb7a2833b8975b4c2/lib/libClp.so (unknown line)
_ZN8ClpModelD1Ev at /home/odow/.julia/artifacts/9cf32528460420740b8ba41eb7a2833b8975b4c2/lib/libClp.so (unknown line)
_ZN10ClpSimplexD1Ev at /home/odow/.julia/artifacts/9cf32528460420740b8ba41eb7a2833b8975b4c2/lib/libClp.so (unknown line)
Clp_deleteModel at /home/odow/.julia/artifacts/9cf32528460420740b8ba41eb7a2833b8975b4c2/lib/libClp.so (unknown line)
trigger at ./REPL[32]:10
unknown function (ip: 0x7f43a9d6b80c)
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2141 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2305
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1631 [inlined]
do_call at /buildworker/worker/package_linux64/build/src/interpreter.c:328
eval_value at /buildworker/worker/package_linux64/build/src/interpreter.c:417
eval_stmt_value at /buildworker/worker/package_linux64/build/src/interpreter.c:368 [inlined]
eval_body at /buildworker/worker/package_linux64/build/src/interpreter.c:778
jl_interpret_toplevel_thunk_callback at /buildworker/worker/package_linux64/build/src/interpreter.c:888
unknown function (ip: 0xfffffffffffffffe)
unknown function (ip: 0x7f43c345da8f)
unknown function (ip: 0xffffffffffffffff)
jl_interpret_toplevel_thunk at /buildworker/worker/package_linux64/build/src/interpreter.c:897
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:814
jl_toplevel_eval_flex at /buildworker/worker/package_linux64/build/src/toplevel.c:764
jl_toplevel_eval_in at /buildworker/worker/package_linux64/build/src/toplevel.c:843
eval at ./boot.jl:330
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2135 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2305
eval_user_input at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/REPL/src/REPL.jl:86
macro expansion at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.3/REPL/src/REPL.jl:118 [inlined]
#26 at ./task.jl:333
_jl_invoke at /buildworker/worker/package_linux64/build/src/gf.c:2135 [inlined]
jl_apply_generic at /buildworker/worker/package_linux64/build/src/gf.c:2305
jl_apply at /buildworker/worker/package_linux64/build/src/julia.h:1631 [inlined]
start_task at /buildworker/worker/package_linux64/build/src/task.c:659
unknown function (ip: 0xffffffffffffffff)
Allocations: 5993989 (Pool: 5992835; Big: 1154); GC: 5
Aborted
odow@antarctic:~$ 
```

ViralBShah:
I guess that should be filed upstream. I should have our versions synced with the homebrew-coinor versions by end of day. It's slow because I have to update each one and register it separately.

odow:
Filed upstream: https://github.com/coin-or/Clp/issues/142.

mtanneau:
Bugs beware, @odow is coming to get you :)

ViralBShah:
1.17.3 is registered. Without ASL and Metis.

odow:
Still getting the invalid pointer on `antarctic` :(

ViralBShah:
Yeah, the code changes from 1.17.3 to 1.17.5 were not substantial. I think it would be best to give upstream time to resolve this, before considering a downgrade of the whole set of libraries further - especially Cbc which is using new versions of everything.

ViralBShah:
I guess one could bisect Clp since we have a very accurate reproducer.

odow:
Yeah, upstream is unlikely to fix without our help. I'll take a shot this weekend unless someone beats me to it.

odow:
Woo! All checks green.

giordano:
You probably want to add a conditional in `deps/build.jl` so that the content of the file is evaluated only if
```julia
haskey(ENV,"JULIA_CLP_LIBRARY_PATH") || VERSION < v"1.3"
```

odow:
Bad news on the mumps front. Doesn't appear to have much of an impact. Still a 2x loss in performance compared with the previous BinaryBuilder release. I'm tempted to class this as "CoinBLAS has some optimizations to make Clp run fast."

These timings are consistent with the ones I did using Clp 1.16.5: https://github.com/JuliaPackaging/Yggdrasil/issues/509#issuecomment-608545018

```julia
# ~/Desktop/bench_clp.jl

import Pkg
if length(ARGS) > 0
    if ARGS[1] == "1.17.6+1"
        Pkg.add(Pkg.PackageSpec(
            name = "Clp_jll", rev = "01c4b8ae7abe4d0bf04910cf5674ee910830477b"
        ))
    elseif ARGS[1] == "1.17.6+0"
        Pkg.add(Pkg.PackageSpec(
            name = "Clp_jll", rev = "44ed7f3781610ad813dc3402e4604326ae779565"
        ))
    end
    Pkg.update()
end
Pkg.status()
using Clp
const MOI = Clp.MOI
download(
    "ftp://ftp.numerical.rl.ac.uk/pub/cuter/netlib/QAP12.SIF",
    "QAP12.mps"
)
function bench()
    mps = MOI.FileFormats.MPS.Model()
    MOI.read_from_file(mps, "QAP12.mps")
    model = Clp.Optimizer()
    MOI.copy_to(model, mps)
    MOI.set(model, MOI.Silent(), true)
    @time MOI.optimize!(model)
end
@time bench()
```


```
Oscars-MBP:Clp oscar$ ~/julia1.3 --project=/tmp/clp ~/Desktop/bench_clp.jl
    Status `/private/tmp/clp/Project.toml`
  [e2554f3b] Clp v0.7.1
 23.901250 seconds
 27.553734 seconds (13.47 M allocations: 633.354 MiB, 1.15% gc time)
Oscars-MBP:Clp oscar$ ~/julia1.3 --project=. ~/Desktop/bench_clp.jl 1.17.6+0
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Updating git-repo `https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git`
 Resolving package versions...
  Updating `~/.julia/dev/Clp/Project.toml`
  [06985876] ↓ Clp_jll v1.17.6+1 #01c4b8a (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git) ⇒ v1.17.6+0 #44ed7f3 (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git)
  Updating `~/.julia/dev/Clp/Manifest.toml`
  [06985876] ↓ Clp_jll v1.17.6+1 #01c4b8a (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git) ⇒ v1.17.6+0 #44ed7f3 (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git)
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Updating git-repo `https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git`
 Resolving package versions...
  Updating `~/.julia/dev/Clp/Project.toml`
 [no changes]
  Updating `~/.julia/dev/Clp/Manifest.toml`
 [no changes]
Project Clp v0.7.1
    Status `~/.julia/dev/Clp/Project.toml`
  [b99e7846] BinaryProvider v0.5.8
  [06985876] Clp_jll v1.17.6+0 #44ed7f3 (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git)
  [b8f27783] MathOptInterface v0.9.13
  [fdba3010] MathProgBase v0.7.8
  [8f399da3] Libdl 
  [37e2e46d] LinearAlgebra 
  [2f01184e] SparseArrays 
 46.027738 seconds
 50.413522 seconds (13.34 M allocations: 627.068 MiB, 0.42% gc time)
Oscars-MBP:Clp oscar$ ~/julia1.3 --project=. ~/Desktop/bench_clp.jl 1.17.6+1
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Updating git-repo `https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git`
 Resolving package versions...
  Updating `~/.julia/dev/Clp/Project.toml`
  [06985876] ↑ Clp_jll v1.17.6+0 #44ed7f3 (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git) ⇒ v1.17.6+1 #01c4b8a (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git)
  Updating `~/.julia/dev/Clp/Manifest.toml`
  [06985876] ↑ Clp_jll v1.17.6+0 #44ed7f3 (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git) ⇒ v1.17.6+1 #01c4b8a (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git)
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Updating git-repo `https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git`
 Resolving package versions...
  Updating `~/.julia/dev/Clp/Project.toml`
 [no changes]
  Updating `~/.julia/dev/Clp/Manifest.toml`
 [no changes]
Project Clp v0.7.1
    Status `~/.julia/dev/Clp/Project.toml`
  [b99e7846] BinaryProvider v0.5.8
  [06985876] Clp_jll v1.17.6+1 #01c4b8a (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git)
  [b8f27783] MathOptInterface v0.9.13
  [fdba3010] MathProgBase v0.7.8
  [8f399da3] Libdl 
  [37e2e46d] LinearAlgebra 
  [2f01184e] SparseArrays 
 47.000191 seconds
 51.442308 seconds (13.34 M allocations: 626.954 MiB, 0.41% gc time)
```

mlubin:
Could you also print the actual time in the solver?

My general impression is that CoinBLAS doesn't do anything smart, nor does Clp use BLAS much at all, so I'm confused by these results. Are the compiler optimization flags the same? 

Also, QAP12 may be an unusual instance, I would try a couple others also.

odow:
I had printing turned on locally. It is time inside the solver (you can see the iterations take noticeably longer). Will test other instances.

ViralBShah:
CoinBLAS literally just packages the reference BLAS. Are we using openblas with only 1 thread? That seems to be important for many Coin-OR libraries.

ViralBShah:
I think the old Builders also had `-DNDEBUG` and I wonder if disabling all debugging makes a huge difference. I can try it out after we look into things a bit more here in the next round.

mlubin:
A quick search through the Clp source shows a pretty liberal use of `assert`. No idea if that could cause a 2x slowdown, however.

ViralBShah:
Yeah that can do it

odow:
2x performance gap holds pretty consistently across instances. For example, on `"ftp://ftp.numerical.rl.ac.uk/pub/cuter/netlib/80BAU3B.SIF"`

```
Oscars-MBP:Clp oscar$ ~/julia1.3 --project=/tmp/clp ~/Desktop/bench_clp.jl
    Status `/private/tmp/clp/Project.toml`
  [e2554f3b] Clp v0.7.1
Coin0506I Presolve 1779 (-483) rows, 8572 (-1227) columns and 18802 (-2200) elements
Clp0006I 0  Obj 495454.62 Primal inf 83504.592 (650) Dual inf 249967.53 (1077)
Clp0006I 110  Obj -1.5738521e+13 Primal inf 1.0640635e+13 (832)
Clp0006I 220  Obj -5.7821056e+11 Primal inf 8.7888851e+11 (1015)
Clp0006I 330  Obj -460880.02 Primal inf 421961.28 (983)
Clp0006I 440  Obj 50199.373 Primal inf 151279.89 (911)
Clp0006I 550  Obj 201455.24 Primal inf 103045.14 (885)
Clp0006I 660  Obj 294304.42 Primal inf 75760.282 (868)
Clp0006I 770  Obj 408741.76 Primal inf 56909.697 (835)
Clp0006I 880  Obj 458784.44 Primal inf 65138.018 (820)
Clp0006I 990  Obj 486771.23 Primal inf 57126.469 (817)
Clp0006I 1100  Obj 533777.08 Primal inf 49571.77 (790)
Clp0006I 1210  Obj 576732.85 Primal inf 80090.654 (780)
Clp0006I 1320  Obj 619384.97 Primal inf 62168.075 (777)
Clp0006I 1430  Obj 653545.23 Primal inf 34004.801 (763)
Clp0006I 1540  Obj 714717.59 Primal inf 46151.088 (751)
Clp0006I 1650  Obj 757672.53 Primal inf 21918.045 (686)
Clp0006I 1760  Obj 787526.67 Primal inf 16891.015 (659)
Clp0006I 1870  Obj 810076.62 Primal inf 20391.899 (653)
Clp0006I 1980  Obj 860401.58 Primal inf 19470.978 (585)
Clp0006I 2090  Obj 891417.6 Primal inf 10094.134 (563)
Clp0006I 2200  Obj 913805.98 Primal inf 9835.4044 (512)
Clp0006I 2310  Obj 926009.94 Primal inf 5813.686 (487)
Clp0006I 2420  Obj 938313.33 Primal inf 3970.137 (453)
Clp0006I 2530  Obj 953457.54 Primal inf 2856.4934 (424)
Clp0006I 2640  Obj 962422 Primal inf 11275.913 (393)
Clp0006I 2750  Obj 970689.75 Primal inf 1224.0817 (350)
Clp0006I 2860  Obj 977424.81 Primal inf 910.74223 (318)
Clp0006I 2970  Obj 983027.01 Primal inf 256.42084 (254)
Clp0006I 3080  Obj 986609.46 Primal inf 73.92618 (209)
Clp0006I 3190  Obj 987165.06 Primal inf 3.4073631 (142)
Clp0006I 3300  Obj 987224.11 Primal inf 0.014131438 (64)
Clp0006I 3380  Obj 987224.19
Clp0000I Optimal - objective value 987224.19
Coin0511I After Postsolve, objective 987224.19, infeasibilities - dual 351.40318 (7), primal 0 (0)
Coin0512I Presolved model was optimal, full model needs cleaning up
Clp0000I Optimal - objective value 987224.19
Clp0032I Optimal objective 987224.1924 - 3380 iterations time 0.152, Presolve 0.04
  0.150701 seconds
  3.800487 seconds (13.55 M allocations: 625.098 MiB, 8.18% gc time)
Oscars-MBP:Clp oscar$ ~/julia1.3 --project=. ~/Desktop/bench_clp.jl 1.17.6+1
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Updating git-repo `https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git`
 Resolving package versions...
  Updating `~/.julia/dev/Clp/Project.toml`
 [no changes]
  Updating `~/.julia/dev/Clp/Manifest.toml`
 [no changes]
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Updating git-repo `https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git`
 Resolving package versions...
  Updating `~/.julia/dev/Clp/Project.toml`
 [no changes]
  Updating `~/.julia/dev/Clp/Manifest.toml`
 [no changes]
Project Clp v0.7.1
    Status `~/.julia/dev/Clp/Project.toml`
  [b99e7846] BinaryProvider v0.5.8
  [06985876] Clp_jll v1.17.6+1 #01c4b8a (https://github.com/JuliaBinaryWrappers/Clp_jll.jl.git)
  [b8f27783] MathOptInterface v0.9.13
  [fdba3010] MathProgBase v0.7.8
  [8f399da3] Libdl 
  [37e2e46d] LinearAlgebra 
  [2f01184e] SparseArrays 
Coin0506I Presolve 1779 (-483) rows, 8572 (-1227) columns and 18802 (-2200) elements
Clp0006I 0  Obj 493096.66 Primal inf 83620.824 (650) Dual inf 249760.54 (1076)
Clp0006I 110  Obj -1.5738521e+13 Primal inf 1.0640635e+13 (832)
Clp0006I 220  Obj -5.7821056e+11 Primal inf 8.7888851e+11 (1015)
Clp0006I 330  Obj -460880.02 Primal inf 421961.28 (983)
Clp0006I 440  Obj 50199.373 Primal inf 151279.89 (911)
Clp0006I 550  Obj 201455.24 Primal inf 103045.14 (885)
Clp0006I 660  Obj 294304.42 Primal inf 75760.282 (868)
Clp0006I 770  Obj 408741.76 Primal inf 56909.697 (835)
Clp0006I 880  Obj 458784.44 Primal inf 65138.018 (820)
Clp0006I 990  Obj 486771.23 Primal inf 57126.469 (817)
Clp0006I 1100  Obj 533777.08 Primal inf 49571.77 (790)
Clp0006I 1210  Obj 576732.85 Primal inf 80090.654 (780)
Clp0006I 1320  Obj 619384.97 Primal inf 62168.075 (777)
Clp0006I 1430  Obj 653545.23 Primal inf 34004.801 (763)
Clp0006I 1540  Obj 714717.59 Primal inf 46151.088 (751)
Clp0006I 1650  Obj 757672.53 Primal inf 21918.045 (686)
Clp0006I 1760  Obj 787526.67 Primal inf 16891.015 (659)
Clp0006I 1870  Obj 810076.62 Primal inf 20391.899 (653)
Clp0006I 1980  Obj 860401.58 Primal inf 19470.978 (585)
Clp0006I 2090  Obj 891417.6 Primal inf 10094.134 (563)
Clp0006I 2200  Obj 913805.98 Primal inf 9835.4044 (512)
Clp0006I 2310  Obj 926009.94 Primal inf 5813.686 (487)
Clp0006I 2420  Obj 938313.33 Primal inf 3970.137 (453)
Clp0006I 2530  Obj 953457.54 Primal inf 2853.2144 (423)
Clp0006I 2640  Obj 962247 Primal inf 5660.8752 (392)
Clp0006I 2750  Obj 970685.22 Primal inf 1234.7429 (353)
Clp0006I 2860  Obj 977351.64 Primal inf 936.45569 (323)
Clp0006I 2970  Obj 983021.28 Primal inf 261.07734 (256)
Clp0006I 3080  Obj 986583.44 Primal inf 80.201399 (211)
Clp0006I 3190  Obj 987165.06 Primal inf 3.4805738 (145)
Clp0006I 3300  Obj 987224.11 Primal inf 0.015285705 (66)
Clp0006I 3382  Obj 987224.19
Clp0000I Optimal - objective value 987224.19
Coin0511I After Postsolve, objective 987224.19, infeasibilities - dual 351.40318 (7), primal 0 (0)
Coin0512I Presolved model was optimal, full model needs cleaning up
Clp0006I 0  Obj 987224.19
Clp0000I Optimal - objective value 987224.19
Clp0032I Optimal objective 987224.1924 - 3382 iterations time 0.352, Presolve 0.07
  0.358659 seconds
  4.204421 seconds (13.42 M allocations: 618.203 MiB, 3.48% gc time)
```

ViralBShah:
New binaries published. Retry?

ViralBShah:
On my laptop, it still takes about 50 seconds with the new binaries.

ViralBShah:
I tried 1.16, and it is even slower.  65 seconds.

ViralBShah:
@giordano @staticfloat I am seeing the binaries produced by [BinaryBuilder](https://github.com/JuliaPackaging/Yggdrasil/blob/master/C/Coin-OR/Clp/build_tarballs.jl) being about half as slow as the [old ClpBuilder](https://github.com/JuliaOpt/ClpBuilder/blob/master/build_tarballs.jl).

Is this likely to be because of architecture specific optimizations in the compiler in the old builder? The other thing the old ClpBuilder did was link everything statically. Can that explain this difference? Any other ideas for why the performance is so different? If anything, we are linking against a fast BLAS in the BB binaries.

ViralBShah:
Yes, you can see the old builder script I linked above which I assume is how the current binaries are produced. It pretty much does not do doing anything like `-Ofast`. Do compilers do something like `-mtune=native` automatically which we are disabling?

@juan-pablo-vielma Any insights you might have here would be valuable.

juan-pablo-vielma:
Took a quick look. 
I see a missing -DCOIN_USE_MUMPS_MPI_H, but maybe it's on the MUMPS builder?
Howe about ASL? Is the new builder using it? I am not sure how Clp uses it though. 
Do you have pairs of logs for problems that take a few minutes? One crazy idea is static v/s dynamic linking.

ViralBShah:
Yes, we do not need the MUMPS MPI thing, because I have already built a sequential mumps and applied their patches. @odow suggested we don't use ASL at all, and hence we removed it. Passes all tests.

Static vs. dynamic linking is the big one right now, but calls to lapack and mumps are chunky enough that it shouldn't matter. Here's the build logs:

https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=3160&view=results

juan-pablo-vielma:
Maybe looking at more solve logs? In the pair that @odow posted above everything is identical up to iteration 2530 and then things change. Not sure if there is a way to get more detailed logs from Clp

ViralBShah:
The other thing that is different is we are using openblas.

ViralBShah:
On mac the new binaries are 2x slower, but on linux they are about 10% slower.

ViralBShah:
@staticfloat helped me debug that the optimizations flags were being clobbered, and `-O3` wasn't being passed. Soon, we should have new builds and hopefully this is solved.

ViralBShah:
Ok, new binaries published. Please try it out. For me, the optimized binaries are twice as fast as the older ones!

```
# With Clp_jll
julia> @time bench()
 17.777809 seconds
 19.452482 seconds (4.72 M allocations: 196.026 MiB, 0.42% gc time)

# With ClpBuilder
julia> @time bench()
 27.312733 seconds
 31.796835 seconds (16.40 M allocations: 766.499 MiB, 0.98% gc time)
```

odow:
Works for me!

odow:
Thank you for your outstanding help. You've probably learnt far more than you wanted to about how Clp works.

ViralBShah:
> Thank you for your outstanding help. You've probably learnt far more than you wanted to about how Clp works.

It's always fun to learn more about numerical software.  And I am glad that these efforts paid off.

ViralBShah:
I think it will be much easier to update the Coin-OR solvers going forward.

PR #78, opened 2020-04-21
Use Clp_jll binaries on Julia pre v1.3 too

blegat:
It's better to use the same binaries accross all Julia versions.
Requires https://github.com/JuliaPackaging/BinaryProvider.jl/pull/192


.travis.yml
```diff
@@ -8,6 +8,9 @@ julia:
   - 1.3
   - 1.4
 
+before_script:
+  - julia -e 'using Pkg; Pkg.add(PackageSpec(name="BinaryProvider", rev="mg/fix-write-deps-isolation"))'
```
odow, line 12: When will this be merged and tagged? Why is this needed only on travis?

blegat: It segfaults when we do `dlopen`. As this is a system call, I guess the failure is os specific and Windows is unaffected. No idea when it is going to be merged.

odow: And this doesn't happen locally? This PR seems dangerous. We know the current version works, at least.

blegat: It happens locally on my computer.

blegat: The failure is well understood, if `isolate=true` is passed, the library is checked with the following line:
https://github.com/JuliaPackaging/BinaryProvider.jl/blob/55aeaea2652c1b43ba882c2f833196e99f10e2db/src/Products.jl#L169
which tries to open the library in a new Julia session.
If `isolate=false`, the library is `dlopen`'d and `dlclose`'d in the current Julia session and `dlclose` segfaults on linux:
https://github.com/JuliaPackaging/BinaryProvider.jl/blob/55aeaea2652c1b43ba882c2f833196e99f10e2db/src/Products.jl#L175
for the `OpenBLAS32` library as Julia also has its own version of this library.
With the current binaries, depending on the solver we either statically link openblas or we use the Julia openblas (which does not work on ArchLinux where the Julia openblas is the system one which does not have the `64_` suffix to the symbols) so we don't have the segfault.

Issue #79, opened 2020-04-24
Default LogLevel is 1 rather than 0 as documented in README


pierre-haessig:
Hello,

Clp.jl's [README §Solver options](https://github.com/JuliaOpt/Clp.jl/blob/master/README.md#solver-options) says that the `LogLevel` is set to zero by default. However, if I run an JuMP optimization such as:

```
model = Model(Clp.Optimizer)
@variable(model, 0 <= x <= 2)
optimize!(model)
```

I get this:

```
Clp3002W Empty problem - 0 rows, 1 columns and 0 elements
Clp0000I Optimal - objective value 0
Clp0032I Optimal objective 0 - 0 iterations time 0.002
```

and indeed `get_optimizer_attribute(model, "LogLevel")` returns `1`.

Perhaps it's a recent change related to the MPB → MOI transition?

I guess either the README or the default `LogLevel` should be updated. I didn't find in the source code where `LogLevel` is set.

best,
Pierre

versions:

- julia 1.4.0
- Clp v0.7.1
- JuMP v0.21.2

mtanneau:
Correct. The default value of `LogLevel` is one. I'll update the README, since the current behavior is consistent with MathOptInterface's [convention](https://www.juliaopt.org/MathOptInterface.jl/stable/apireference/#MathOptInterface.Silent).

Thanks for reporting this!

odow:
Closed by #80

PR #80, opened 2020-04-24
Update README

mtanneau:
Following #79 , I think a few updates to the README are due by now:

* [x] Default values of Clp options
* [ ] New JuMP syntax
* [ ] Mention that the `Clp_jll` binaries are used


odow:
Other PRs incoming.

PR #82, opened 2020-04-29
Typos in load_quadratic_objective corrected

MassimoDeMauri:
Hi,
I just noticed that a couple of typos in both the load_quadratic_objective functions.
It would be nice if you could include this small correction in the main repository.

Have a nice day,
Massimo.


odow:
The quadratic interface to Clp is not tested. I also don't know how stable/reliable it is.

MassimoDeMauri:
Thanks a lot, I will test it in the next future and come back to you with any fix/update.

odow:
Do you have an opinion on this? https://github.com/JuliaOpt/Clp.jl/pull/87

PR #83, opened 2020-05-03
Switch to loadProblem instead of incremental building and tidy MOI wrapper

odow:
Notably, on the `test_1.mps` problem from this post, https://discourse.julialang.org/t/cbc-and-clp-performance/38505/6, we go from not even finishing the load in 10 minutes, to 40 seconds.

```Julia
julia> model
A JuMP Model
Minimization problem with:
Variables: 574369
Objective function type: VariableRef
`GenericAffExpr{Float64,VariableRef}`-in-`MathOptInterface.EqualTo{Float64}`: 156985 constraints
`GenericAffExpr{Float64,VariableRef}`-in-`MathOptInterface.GreaterThan{Float64}`: 401756 constraints
`GenericAffExpr{Float64,VariableRef}`-in-`MathOptInterface.LessThan{Float64}`: 319172 constraints
`VariableRef`-in-`MathOptInterface.GreaterThan{Float64}`: 566929 constraints
Model mode: AUTOMATIC
CachingOptimizer state: ATTACHED_OPTIMIZER
Solver name: Clp
```

Here's the last bit of the `optimize!` log:
```
Clp0000I Optimal - objective value 1.8732729e+09
Coin0511I After Postsolve, objective 1.8732729e+09, infeasibilities - dual 28018158 (5423), primal 0.00073130792 (15)
Coin0512I Presolved model was optimal, full model needs cleaning up
Clp0006I 0  Obj 1.8732729e+09 Primal inf 8.9651245e-06 (14) Dual inf 1.3999997e+09 (14)
Clp0006I 211  Obj 1.8732729e+09 Primal inf 2.4652618e-06 (3) Dual inf 2.9999989e+08 (3)
Clp0029I End of values pass after 331 iterations
Clp0006I 331  Obj 1.8732729e+09
Clp0000I Optimal - objective value 1.8732729e+09
Clp0032I Optimal objective 1873272903 - 461755 iterations time 841.682, Presolve 25.13
882.713930 seconds (115.69 M allocations: 4.828 GiB, 0.49% gc time)
```

### DIFF
--- a/src/MOIWrapper.jl
+++ b/src/MOIWrapper.jl
@@ -136,8 +136,8 @@ MOI.supports(::Optimizer, ::MOI.NumberOfThreads) = false
 # ========================================
 
 function MOI.supports_constraint(
-    ::Optimizer, ::Type{MOI.ScalarAffineFunction{Float64}}, ::Type{S}
-) where {S<:SCALAR_SETS}
+    ::Optimizer, ::Type{MOI.ScalarAffineFunction{Float64}}, ::Type{<:SCALAR_SETS}
+)
     return true
 end
 
@@ -336,7 +336,6 @@ function MOI.get(model::Optimizer, ::MOI.RawStatusString)
 end
 
 function MOI.get(model::Optimizer, ::MOI.ResultCount)
-    st = MOI.get(model, MOI.TerminationStatus())
     if Clp.primal_feasible(model.inner)
         return 1
     elseif Clp.dual_feasible(model.inner)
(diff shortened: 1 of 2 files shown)


mlubin:
For comparison, how long does Clp take if you call it from the command line with the mps file?

odow:
I don't have an identical binary to the one we ship, but for 1.17.5 on homebrew, I got 600 seconds.  But it had a different number of iterations, and more time spent in presolve.

We're in the right ballpark at least.

odow:
I found a heap of other performance issues once I dug in. The whole wrapper could do with a good re-write, strip out the MPB stuff, and build the C interface using Clang.

src/MOIWrapper.jl
```diff
@@ -147,174 +134,146 @@ MOI.supports(::Optimizer, ::MOI.NumberOfThreads) = false
...
 # ========================================
-MOI.supports_constraint(::Optimizer,
-    ::Type{MOI.ScalarAffineFunction{Float64}}, ::Type{<:SCALAR_SETS}
-) = true
 
-MOI.supports_constraint(::Optimizer,
-    ::Type{MOI.SingleVariable}, ::Type{<:SCALAR_SETS}
-) = true
+function MOI.supports_constraint(
+    ::Optimizer, ::Type{MOI.ScalarAffineFunction{Float64}}, ::Type{S}
```
blegat, line 139: Why creating this variable S? It's not recommended by Julia style guide : https://docs.julialang.org/en/v1/manual/style-guide/index.html#Don't-use-unnecessary-static-parameters-1

PR #87, opened 2020-05-07
Add Clang wrapper. Remove MPB

### LINKED ISSUE #84: Don't query entire solution vector every time
This is really bad for performance, because we copy the entire vector into Julia every call.
https://github.com/JuliaOpt/Clp.jl/blob/a812d4f00b6b4c297220ce375ce477c1e78ff073/src/MOIWrapper.jl#L430-L431

We should use a function like this instead:
```Julia
function getColSolution(model::ClpModel, column::Int)
    p = Clp.ClpCInterface.@clp_ccall(
        getColSolution, Ptr{Float64}, (Ptr{Cvoid},), model.p
    )
    return unsafe_wrap(Array, p, (Clp.get_num_cols(model),))[column]
end
```

odow:
When I was fixing up the MOI wrapper the other day, I say numerous inefficiencies in the C API. I started to look through, when I realized it would actually be pretty easy to use Clang to automatically create the wrapper, especially now that very little of the MOI actually is C calls.

Here's the result.

The dodgiest things we could discuss are:
1. <s>Under what conditions to allow user-installed libraries to deviate from the jll version</s>
2. <s>[How to (un)safely unwrap the solution arrays provided by Clp](https://github.com/JuliaOpt/Clp.jl/compare/od/clang?expand=1#diff-bc57bef498a950328f785a44f62179fcR378-R391). One complication is that it appears the infeasibility and unbounded rays are actually malloc'd and responsibility for freeing them is on our end.</s> I've convinced myself we're doing this correctly.

<s>Edit: I should add: passes tests locally on Mac 1.1 and 1.3. No idea on other platforms.</s>Tests pass.

MBP is dead. Long live MOI.

Closes #84


mtanneau:
> Under what conditions to allow user-installed libraries to deviate from the jll version

What issues may happen if someone doesn't use the jll version?
Wrong C functions? Or can we safely say "if you don't change the C API, you're fine"?

odow:
The wrapper is built for 1.17.6, so presumably we're okay for any `1.17.x`? The only issue is when Clp changes things like `maximumIterations` to `Clp_maximumIterations`.

odow:
Looking at the diff, we're safe to use 1.17.2:
https://github.com/coin-or/Clp/compare/releases/1.17.2...releases/1.17.6#diff-20bd77190b025536c61a52f8e088d14b

In 1.17.1, the added some new methods like `Clp_writeMPS`.

mlubin:
What about a helpful error message if someone tries to call `Clp.ClpSolver()`?

src/Clp.jl
```diff
@@ -31,4 +31,12 @@ end
 
 include("moi/MOI_wrapper.jl")
 
+# TODO(odow): remove at Clp.jl v1.0.0.
+function ClpSolver(args...; kwargs...)
+    error(
+        "`ClpSolver` is no longer supported. Upgrade to JuMP 0.20 or greater," *
```
mlubin, line 37: JuMP 0.19 or greater? Also not everyone calling ClpSolver() is necessarily using JuMP. They could be using `linprog` or another package that uses MPB.

odow: How about ````"`ClpSolver` is not longer supported. If you are using JuMP, upgrade to the latest version and use `Clp.Optimizer` instead. If you are using MathProgBase (e.g., via `lingprog`), you will need to upgrade to MathOptInterface (https://github.com/JuliaOpt/MathOptInterface.jl)."````

I'm guessing the number of users that are going to hit this is very small.

mlubin: Sounds good. The number is small but almost definitely nonzero.

mtanneau:
> The only issue is when Clp changes things like maximumIterations to Clp_maximumIterations

This one did change (See https://github.com/coin-or/Clp/issues/141), though it's not in the latest release yet.

odow:
Yes, it will be in a 1.18 release.

odow:
Objections to merging?

mtanneau:
No.

mlubin:
Go for it

PR #92, opened 2020-05-27
Fix missing argument when calling setters from MOI

### LINKED ISSUE #91: MethodError when setting PresolveType and SolveType through MOI
The following used to work but now results in a `MethodError`:

```julia
import Clp
import MathOptInterface
const MOI = MathOptInterface

clp = Clp.Optimizer()

MOI.set(clp, MOI.RawParameter("SolveType", 4))     # errors
MOI.set(clp, MOI.RawParameter("PresolveType", 1))  # also errors
```
I did not see any other other case where the following 

Error message:
```
ERROR: MethodError: no method matching ClpSolve_setSolveType(::Ptr{Nothing}, ::Int64)
Closest candidates are:
  ClpSolve_setSolveType(::Any, ::Any, ::Any) at /mnt/c/Users/mathi/Git/Clp.jl/src/gen/libclp_api.jl:593
Stacktrace:
 [1] set(::Clp.Optimizer, ::MathOptInterface.RawParameter, ::Int64) at /mnt/c/Users/mathi/Git/Clp.jl/src/MOI_wrapper/MOI_wrapper.jl:101
 [2] top-level scope at REPL[5]:1
```

There are at least 3 ways of fixing this. PR incoming.

mtanneau:
Fixes #91.

According to the [Clp C Interface](https://github.com/coin-or/Clp/blob/a9bd29ba44d0543da31adf73b0603886d312e457/src/Clp_C_Interface.h#L488-L506), setting `extraInfo = -1` results in default behavior.

I can see 3 ways of fixing this:
1. Add a default value to `ClpSolve_setSolveType` as follows:
    ```julia
    ClpSolve_setSolveType(arg1, method, extraInfo=Cint(-1))
    ```
    Pro: very short. Con: that file is generated automatically, so we probably don't want to modify it

2. (the one implemented here): extend the two setters to add the default value
   Pro: fixed the bug. Con: extends a function of the C API.

3. "hack" the `SOLVE_OPTION_MAP` dict as follows:
   ```julia
    :SolveType => (ClpSolve_getSolveType, (arg1, method) -> ClpSolve_setSolveType(arg1, method, -1))
    ```
    Pro: compact. Con: Not great for error messages.

### DIFF
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -98,7 +98,15 @@ function MOI.set(model::Optimizer, param::MOI.RawParameter, value)
         push!(model.options_set, key)
         CLP_OPTION_MAP[key][2](model.inner, value)
     elseif haskey(SOLVE_OPTION_MAP, key)
-        SOLVE_OPTION_MAP[key][2](model.solver_options, value)
+        # Setters for `SolveType` and `PresolveType` are handled separately,
+        # as they have a third `extraInfo` argument
+        if key == :SolveType
+            ClpSolve_setSolveType(model.solver_options, value, -1)
+        elseif key == :PresolveType
+            ClpSolve_setPresolveType(model.solver_options, value, -1)
+        else
+            SOLVE_OPTION_MAP[key][2](model.solver_options, value)
+        end
     else
         throw(MOI.UnsupportedAttribute(param))
     end
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -95,13 +95,13 @@ MOI.supports(::Optimizer, param::MOI.RawParameter) = true
 function MOI.set(model::Optimizer, param::MOI.RawParameter, value)
     key = Symbol(param.name)
     
-    # Setters for `SolveType` and `PresolveType` are handled separately,
-    # as they have a third `extraInfo` argument
+    # Setters for `SolveType`, `PresolveType` and `InfeasibleReturn`
+    # are handled separately, see https://github.com/jump-dev/Clp.jl/issues/91
     if haskey(CLP_OPTION_MAP, key)
         push!(model.options_set, key)
         CLP_OPTION_MAP[key][2](model.inner, value)
     elseif key == :SolveType
-            ClpSolve_setSolveType(model.solver_options, value, -1)
+        ClpSolve_setSolveType(model.solver_options, value, -1)
     elseif key == :PresolveType
         ClpSolve_setPresolveType(model.solver_options, value, -1)
     elseif key == :InfeasibleReturn
--- a/src/MOI_wrapper/MOI_wrapper.jl
+++ b/src/MOI_wrapper/MOI_wrapper.jl
@@ -94,19 +94,18 @@ MOI.supports(::Optimizer, param::MOI.RawParameter) = true
 
 function MOI.set(model::Optimizer, param::MOI.RawParameter, value)
     key = Symbol(param.name)
+    
+    # Setters for `SolveType` and `PresolveType` are handled separately,
+    # as they have a third `extraInfo` argument
     if haskey(CLP_OPTION_MAP, key)
         push!(model.options_set, key)
         CLP_OPTION_MAP[key][2](model.inner, value)
-    elseif haskey(SOLVE_OPTION_MAP, key)
-        # Setters for `SolveType` and `PresolveType` are handled separately,
-        # as they have a third `extraInfo` argument
-        if key == :SolveType
+    elseif key == :SolveType
             ClpSolve_setSolveType(model.solver_options, value, -1)
-        elseif key == :PresolveType
-            ClpSolve_setPresolveType(model.solver_options, value, -1)
-        else
-            SOLVE_OPTION_MAP[key][2](model.solver_options, value)
-        end
+    elseif key == :PresolveType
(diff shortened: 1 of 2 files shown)


blegat:
1. We discussed this in the JuMP-dev call last week and concluded that we should not modify the generated file.
2. Sounds good but it is modifying the C API so it's visible from a user wanting to use the C API from Julia. We need to decide if we want to allow that in the wrappers in general. Another option could be to create a `_...` function.
3. Why is it worse for error messages ? Because it's using an anonymous function which is less clear in backtraces ?

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -27,6 +27,11 @@ const CLP_OPTION_MAP = Dict(
     :Algorithm => (Clp_algorithm, Clp_setAlgorithm)
 )
 
+# These two setters have a third argument `extraInfo`, so we extend them
+# to avoid errors when setting the corresponding `MOI.RawAttribute`s
+ClpSolve_setPresolveType(arg1, amount) = ClpSolve_setPresolveType(arg1, amount, -1)
+ClpSolve_setSolveType(arg1, method) = ClpSolve_setSolveType(arg1, method, -1)
```
odow, line 33: I am very strongly against adding methods like this.

We should fix the handling of the `RawParameter` to account for the need to pass multiple arguments.

```Julia
function MOI.set(model, p::MOI.RawParameter, args...)
    # ... some stuff ...
    ClpSolve_setSolveType(model.inner, args...)
end

Comment (odow) [reply src/MOI_wrapper/MOI_wrapper.jl:33]: Or we do
```Julia
function MOI.set(model, p::MOI.RawParameter, value)
    if p.name == :SolveType
        ClpSolve_setSolveType(model.solver_options, value, -1)
    else
        error("Parameter $p not valid")
    end
end
```

odow: These `if` statements are probably nicer than the dictionaries

mtanneau:
Regardless of implementation, the question is:
do we want users to write this
```julia
MOI.set(clp, MOI.RawParameter("SolveType"), 4)  # extraInfo hidden
```
or this
```julia
MOI.set(clp, MOI.RawParameter("SolveType"), 4, -1) # extraInfo exposed
MOI.set(clp, MOI.RawParameter("SolveType"), (4, -1)) # also possible
```

The first option hides the `extraInfo` from the MOI interface. I think the loss of functionality is acceptable. It does require the MOI wrapper to intercept the `SolveType` and `PresolveType` cases and handle them explicitly. There are only two such cases which should be tractable.
Finally, it is backward-compatible.

The second option puts the burden on the user to pass the correct value for `extraInfo`.
Note that `RawParameter`s are used in the `Optimizer` constructor [here](https://github.com/JuliaOpt/Clp.jl/blob/dfb2c475788f86dafb7f7ce2849c1957ebc4a638/src/MOI_wrapper/MOI_wrapper.jl#L55), which also raises the question of how to do this
```julia
clp = Clp.Optimizer(SolveType=4)
```
That is why I suggested passing a tuple `(4, -1)` instead of two arguments above.

odow:
I think hiding the `extraInfo` is fine. Users can always access the full C API if they want full control.

The the constructors: we want to drop this at some point in favor of `RawParameter` so that the behavior is standardized across solvers (some of which don't support kwargs). Now might be a good point to throw an error if people pass arguments.

mtanneau:
> I think hiding the extraInfo is fine. Users can always access the full C API if they want full control.

:+1:

Removing keyword arguments from the constructor would be breaking; I can include it here or open a separate PR.

odow:
> Removing keyword arguments from the constructor would be breaking; I can include it here or open a separate PR.

Separate PR. We probably just want to throw a warning for now. Then remove in a future release (e.g., Clp 1.0).

mtanneau:
> What do the getters look like?

Getters are not affected since they don't have any `extraInfo` argument. I did not change anything on that side.

mtanneau:
For completeness, there _are_ some options for which getters/setters require multiple arguments: [`SpecialOption`](https://github.com/coin-or/Clp/blob/a9bd29ba44d0543da31adf73b0603886d312e457/src/Clp_C_Interface.h#L485-L486) and [`ExtraInfo`]().

I have no idea what they do (`SpecialOptions` is barely documented) nor have I ever used them. They are also currently not exposed at the MOI level, and I think it's OK to keep it that way until someone asks for it.

src/MOI_wrapper/MOI_wrapper.jl
```diff
@@ -94,11 +94,18 @@ MOI.supports(::Optimizer, param::MOI.RawParameter) = true
...
     if haskey(CLP_OPTION_MAP, key)
         push!(model.options_set, key)
         CLP_OPTION_MAP[key][2](model.inner, value)
-    elseif haskey(SOLVE_OPTION_MAP, key)
-        SOLVE_OPTION_MAP[key][2](model.solver_options, value)
+    elseif key == :SolveType
+            ClpSolve_setSolveType(model.solver_options, value, -1)
+    elseif key == :PresolveType
+        ClpSolve_setPresolveType(model.solver_options, value, -1)
+    elseif key == :InfeasibleReturn
```
blegat, line 107: I would have expected `elseif haskey(SOLVE_OPTION_MAP, key)` to be consistent with the comment `Setters for `SolveType` and `PresolveType` are handled separately`. Either update this line or the comment.

mtanneau: Comment updated. I also fixed the indent.

PR #98, opened 2020-10-09
Update README.md

johnzigla:
As discussed on discourse, some of the documentation is out of date: https://discourse.julialang.org/t/documentation-of-clp-jl/48005

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,5 +73,6 @@ The following options are the most useful (and well documented):
 | `MaximumSeconds` | `-1.0` | Terminate after this many seconds have passed. A negative value means no time limit |
 | `LogLevel` | `1` | Set to 1, 2, 3, or 4 for increasing output. Set to `0` to disable output |
 | `PresolveType` | `0` | Set to 1 to disable presolve |
-| `SolveType` | `5` | Solution method: dual simplex (`0`), primal simplex (`1`), sprint (`2`), barrier with crossover (`3`), barrier without crossover (`4`), automatic (`5`) |
+| `Algorithm` | `5` | Solution method: dual simplex (`0`), primal simplex (`1`), sprint (`2`), barrier with crossover (`3`), barrier without crossover (`4`), automatic (`5`) | 
+| `Scaling` |   | off (`0`), equilibrium (`1`), geometric (`2`), auto (`3`), auto-but-as-initialSolve-in-bab (`4`) |
 | `InfeasibleReturn` | `0` | Set to 1 to return as soon as the problem is found to be infeasible (by default, an infeasibility proof is computed as well) |


README.md
```diff
@@ -73,5 +73,6 @@ The following options are the most useful (and well documented):
 | `MaximumSeconds` | `-1.0` | Terminate after this many seconds have passed. A negative value means no time limit |
 | `LogLevel` | `1` | Set to 1, 2, 3, or 4 for increasing output. Set to `0` to disable output |
 | `PresolveType` | `0` | Set to 1 to disable presolve |
-| `SolveType` | `5` | Solution method: dual simplex (`0`), primal simplex (`1`), sprint (`2`), barrier with crossover (`3`), barrier without crossover (`4`), automatic (`5`) |
+| `Algorithm` | `5` | Solution method: dual simplex (`0`), primal simplex (`1`), sprint (`2`), barrier with crossover (`3`), barrier without crossover (`4`), automatic (`5`) | 
+| `Scaling` |   | off (`0`), equilibrium (`1`), geometric (`2`), auto (`3`), auto-but-as-initialSolve-in-bab (`4`) |
```
mlubin, line 77: The default value for `Scaling` appears to be missing from the table.

johnzigla: The default value for Scaling is likely to be `3`, but it is not stated in the documentation, so I left it blank.

odow: The default is `3`:
```Julia
julia> using Clp; const MOI = Clp.MOI;

julia> model = Clp.Optimizer()
Clp.Optimizer

julia> MOI.get(model, MOI.RawParameter("Scaling"))
3
```

README.md
```diff
@@ -73,5 +73,6 @@ The following options are the most useful (and well documented):
 | `MaximumSeconds` | `-1.0` | Terminate after this many seconds have passed. A negative value means no time limit |
 | `LogLevel` | `1` | Set to 1, 2, 3, or 4 for increasing output. Set to `0` to disable output |
 | `PresolveType` | `0` | Set to 1 to disable presolve |
-| `SolveType` | `5` | Solution method: dual simplex (`0`), primal simplex (`1`), sprint (`2`), barrier with crossover (`3`), barrier without crossover (`4`), automatic (`5`) |
+| `Algorithm` | `5` | Solution method: dual simplex (`0`), primal simplex (`1`), sprint (`2`), barrier with crossover (`3`), barrier without crossover (`4`), automatic (`5`) |
```
odow, line 76: Where is the documentation for this? It should be `SolveType`.
https://github.com/coin-or/Clp/blob/756ddd3ed813eb1fa8b2d1b4fe813e6a4d7aa1eb/Clp/src/Clp_C_Interface.h#L485-L494

https://github.com/jump-dev/Clp.jl/blob/b271b17fe4aa4a091d7ac3b53084d44f09d8d38a/src/MOI_wrapper/MOI_wrapper.jl#L103-L104

johnzigla: `SolveType` used to work in the past, but now it gives me an error message. In the documentation for Clp.jl under **Using with JuMP**, the following syntax is used:```set_optimizer_attribute(model, "Algorithm", 4)``` https://github.com/jump-dev/Clp.jl

mlubin: What's the error message when you use "SolveType"?

odow: I can't reproduce the error?
```Julia
julia> model = Model(Clp.Optimizer)
A JuMP Model
Feasibility problem with:
Variables: 0
Model mode: AUTOMATIC
CachingOptimizer state: EMPTY_OPTIMIZER
Solver name: Clp

julia> set_optimizer_attribute(model, "SolveType", 4)
4

julia> optimize!(model)
Clp3002W Empty problem - 0 rows, 0 columns and 0 elements
```

johnzigla: The error message was something like "attribute not supported", and I am certain that I was using the correct syntax. I have experienced similar problems with other packages in the past where suddenly some features stopped working. 
After a complete reinstall of Julia, I can use `SolveType` without any problems. I have tested the `Algorithm' option and it does not seem to change the solution method (ie. simplex, barrier, etc), but `SolveType` does

odow: Thanks for the follow-up. I've refactored our option handling here: https://github.com/jump-dev/Clp.jl/pull/99. It also adds a docstring to `Clp.Optimizer` which mentions the full list of supported options.

PR #109, opened 2020-11-04
Switch to GH actions

odow:
Let's see if this works.


mlubin:
Nice! The other complex part that's not tested here is pushing documentation from documenter.

odow:
Documenter looks like it is okay: https://juliadocs.github.io/Documenter.jl/stable/man/hosting/#GitHub-Actions

jd-lara:
@mlubin we do it this way if it helps 
https://github.com/NREL-SIIP/PowerSystems.jl/blob/master/.github/workflows/docs.yml

.github/workflows/ci.yml
```diff
@@ -0,0 +1,43 @@
...
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        version: ['1.0', '1']  # Test against LTS and current minor release
+        os: [ubuntu-latest, macOS-latest, windows-latest]
+        arch: [x64]
+        include:
+          # Also test against Windows 32-bit on LTS.
+          - version: '1.0'
```
jd-lara, line 19: Is it worth testing against windows 32bit considering it is the second tier in Julia support?

odow: This is what we were doing previously, so I kept it. This is where having system information of who is installing each package would be useful...

mlubin: 32-bit linux would probably be more useful to test than 32-bit windows, but they provide about the same coverage of the code

mlubin (approved):

mlubin:
@odow Do you have the access you need to delete the appveyor project?

odow:
> Do you have the access you need to delete the appveyor project?

Yes. I'll merge this now and check it passes on the `master` branch.

PR #117, opened 2021-06-26
Use default_cache

blegat:
Requires https://github.com/jump-dev/MathOptInterface.jl/pull/1381
Requires https://github.com/jump-dev/MathOptInterface.jl/pull/1412


odow:
Closing in favor of #130

PR #119, opened 2021-08-28
Update wrapper generator to Clang.jl master

melonedo:
On next release, Clang.jl will release a new generator format, this updates the wrapper generator to the new format, and makes uses of JLL packages instead of manually copying headers. Compared using https://github.com/melonedo/WrapperComparator, the new wrapper will be identical to the old one.

cc @gnimuc.


gen/generate.toml
```diff
@@ -0,0 +1,23 @@
+[general]
+library_name = "libClp"
+output_file_path = "../lib/libClp.jl"
+# module_name = "libclp"
```
odow, line 4: Why are some of these commented? Some of the options are pretty opaque. For example how do we know to use `smart_de_anonymize = true`?

lib/libClp.jl
```diff
@@ -1,42 +1,89 @@
...
-
-# Julia wrapper for header: Clp_C_Interface.h
-# Automatically generated using Clang.jl
+const Clp_Solve = Cvoid
 
+"""
+    Clp_Version()
 
+Clp library version number as string. 
+"""
```
odow, line 7: Woo! Having the docstrings is nice.

lib/libClp.jl
```diff
@@ -341,82 +668,186 @@ function Clp_logLevel(model)
...
+    Clp_initialSolve(model)
+
+General solve algorithm which can do presolve. See ClpSolve.hpp for options
+"""
 function Clp_initialSolve(model)
     ccall((:Clp_initialSolve, libClp), Cint, (Ptr{Clp_Simplex},), model)
 end
 
-function Clp_initialSolveWithOptions(model, arg1)
```
odow, line 368: This is the only code change. I guess this was an upstream change?

odow (approved):
LGTM

gen/generate.toml
melonedo: You can check [Clang's own generator.toml](https://github.com/JuliaInterop/Clang.jl/blob/master/gen/generator.toml) for reference. They are commented because I tested them but end up disabling them.

lib/libClp.jl
melonedo: Yes, numbering unnamed arguments by position is more reasonable.
See https://github.com/JuliaGeo/GDAL.jl/pull/120#issuecomment-895889617

odow: Great!

odow:
Thanks! 

As a side note: why'd you pick Clp? Just for test cases for Clang? Or are you interested in optimization? I ask because there are a lot of other solvers in `jump-dev` that have similar wrappers. But upstream is infrequently updated so there's no rush to update them as well.

melonedo:
I update them because we are preparing a new Clang.jl release as in JuliaInterop/Clang.jl#299. Also this is part of my [OSPP](https://summer.iscas.ac.cn/#/?lang=en) project.
Clp is picked because I found its wrapper is not fully automatic.
https://github.com/jump-dev/Clp.jl/blob/9a0b9d143208b38e649525bc3d5516d8879845d5/scripts/clang.jl#L1-L10

odow:
Very cool. I'll comment in the other issue, but don't let the `jump-dev` repos block the release. We can update them as needed.

PR #130, opened 2022-02-22
Implement MOI.default_cache

odow:
Requires https://github.com/jump-dev/MathOptInterface.jl/pull/1760

Replaces #117


test/MOI_wrapper.jl
```diff
@@ -27,12 +27,10 @@ function test_supports_default_copy_to()
 end
 
 function test_runtests()
-    model = MOI.Bridges.full_bridge_optimizer(
-        MOI.Utilities.CachingOptimizer(
-            MOI.Utilities.UniversalFallback(MOI.Utilities.Model{Float64}()),
-            Clp.Optimizer(),
-        ),
-        Float64,
```
blegat, line 35: Add a test here that `model.optimizer.cache` is of the right type

blegat (approved):

PR #146, opened 2023-08-16
Fix support for MOI.TimeLimitSec

odow:
x-ref https://github.com/jump-dev/MathOptInterface.jl/pull/2256


odow:
cc @matbesancon there are a bunch of solvers like this that had slightly varying implementations of TimeLimitSec. This PR is arguably a bug fix, but it is potentially breaking in that querying TimeLimitSec will now return `nothing` when it used to return `-1.0`. Do you think that's okay?

matbesancon:
I would say that's fine. There are always problems with getting unset attribute values, because it often requires the wrapper to store what has been set or not, which is quite cumbersome. 
In this case one could argue that code that breaks was relying on incorrect solver behavior / UB in some way

PR #150, opened 2024-07-24
Load an LP64 backend the new version of Clp_jll.jl

amontoison:


amontoison:
@odow 
Can you trigger this CI builds?

odow:
I guess this is needed, but it is unrelated to the aarch64-apple stuff.

amontoison:
> I guess this is needed, but it is unrelated to the aarch64-apple stuff.

We should try to use a version of `Clp` compiled in debug mode but I highly suspect that the issue is upstream.

odow:
See https://github.com/jump-dev/Clp.jl/issues/131#issuecomment-2235172720

odow (approved):

amontoison:
> I guess this is needed, but it is unrelated to the aarch64-apple stuff.

I'm 99% sure that they never tested the compilation of Clp with `Clang` on Apple Silicon and always tested with `gcc`.
`gcc` was needed during a very long time because of incomplete support of OpenMP with the `Clang` compiler.

